PR #3, opened 2014-12-01
Create a custom watering animation (and wifi icon pixel fix)

ifreislich:
Create a custom watering animation.
Put back the removed pixel in the wifi icon with the 'x'.
Rearrange the wifi icon definitions so they're created in numerical
  order (foolish consistency).


rayshobby:
After reviewing we've decided not to adopt the icon pixel fix. Thanks.

PR #8, opened 2015-04-18
I2C LCD compile option for RPi and BBB

denisfrench:
Hello Ray, Samer et al;

I found a port of the Arduino LiquidCrystal_I2C library for the CubieBoard, and thought it would simplify the process of making the Arduino OpenSprinkler LCD display code cross-compatible with the RPi/BBB.

So I forked it (https://github.com/denisfrench/LiquidCrystal_I2C_RPi) and tinkered a bit to make sure it worked on the RPi, which it did without any modification. I have since changed a couple of #defines that were conflicting with OpenSprinklerGen2.

I'm deliberately using an LCD backpack compatible with Martin Phirt's circuit (http://www.pihrt.com/elektronika/258-moje-rapsberry-pi-i2c-lcd-16x2), as this seems the defacto standard for OSPi/OSPy.

I've done my best to separate out the LCD code from the Arduino push-button UI code using compile-time #defines. I have not tested on Arduino, but it still appears to work fine on OSPi with the standard -DOSPI switch.

Compiling with the -DRPIBBBLCD switch reveals an incompatibility between this library and the original Arduino LiquidCrystal library in that it doesn't seem to like it when you set the base of the value to be printed to the LCD.

The internal workings of the library are still several levels beyond where my learning is at presently, so any input would be greatly appreciated.

There are also outstanding differences between the Arduino and C implementation of their Time libraries which should not be too difficult to solve, but I have not looked into as yet.

Thanks for all your efforts on this so far. I have no idea how good this is as a sprinkler system, but it sure makes a great controller for my pool pump which uses dry contacts to set the speed :-)

Best regards,
Denis.

### DIFF
--- a/I2CDeviceLCD.cpp
+++ b/I2CDeviceLCD.cpp
@@ -1,0 +1,162 @@
+// -------------------------------------------------------------------------------------------
+// Port for CubieBoard a library LiquidCrystal version 1.2.1 to drive the PCF8574* I2C
+//
+// @version API 1.0.0
+//
+// @author Kabanov Andrew - andrew3007@kabit.ru
+// -------------------------------------------------------------------------------------------
+
+#include "I2CDeviceLCD.h"
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <sys/ioctl.h>
+
+#include <stdio.h>
+#include <unistd.h>
+#include <string.h>
+#include <errno.h>
+
+#include <linux/i2c-dev.h>
+#include <linux/limits.h>
+
+// ---------------------------------------------------------------------------------------------------------------
+
+I2CDeviceLCD::I2CDeviceLCD()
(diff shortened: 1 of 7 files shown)


salbahra:
Hi Denis,

Thank you very much for this contribution. We are quite busy getting ready for 2.1.4 release however we will review this after that and get it merged in if everything looks good.

Thank you!

adrian78666:
Could you please help me getting the lcd to work with my ospi? Thank you.

green-striped-gecko:
Hi,
is there any time line if the pull request (adding and LCD) will be added to the firmware in the coming future. If not I will go towards the workaround of stanoba. Just wanted to make sure if it is worth the effort.
Thanks

rayshobby:
Thanks for the PR. I know this is way delayed, but the LCD and button support for RPi have been added to PR #303 and will be merged to master once it's tested and reviewed.

PR #12, opened 2015-05-02
Switch connected to rain sensor terminals turns on/off first program

denisfrench:
I thought this might be useful to someone else out there.

I've connected a simple switch on a length of bell wire to the rain sensor terminals to enable manual running of the first program by remote.

The functionality is enabled with the -DRAINPGM1 compile switch.

The compile switch also disables normal rain sensor operation. This could probably be done more elegantly through the UI, but this method is fine for my purpose.

Best regards,
Denis.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -593,6 +593,13 @@ void check_weather() {
 }
 
 void turn_off_station(byte sid, ulong curr_time) {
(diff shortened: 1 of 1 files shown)


rayshobby:
This will be addressed in the next firmware release as an additional function of sensor port.

PR #18, opened 2015-05-26
Fix expansion board detection (Conflict with RF_DATA)

ifreislich:
The RF Data pin is A3 (27), not A4 (28) as defined in defines.h.
This causes the initialization of the RF pin to pull the EXP_SENSE
pin to 0V breaking detection of expansion boards.

Additionally, the PIN_EXP_SENSE is now initialized to INPUT.

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -471,6 +471,9 @@ void OpenSprinkler::begin() {
   lcd.createChar(3, lcd_rain_char);
   lcd.createChar(4, lcd_connect_char);
 
+  // Expansion detection
+  pinMode(PIN_EXP_SENSE, INPUT);
+
   // set button pins
   // enable internal pullup
   pinMode(PIN_BUTTON_1, INPUT);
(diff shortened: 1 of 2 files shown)


rayshobby:
I am not sure about this request: pin A3 is also digital pin number 28. Note that the digital pin numbers for A0, A1, A2, A3, ... are 31, 30, 29, 28 and so on.
https://github.com/rayshobby/opensprinkler/blob/master/OpenSprinkler%20Controller/software/arduino-code/hardware/aopensprinkler/variants/standard/pins_arduino.h#L66
You may have assumed them in the reverse order.

ifreislich:
OK I should have looked in your included pins_arduino.h.  Seems the confusion comes from the lack of official support for the ATMega644 and ATMega1284 by the Arduino crowd.  It appears that there are two competing "standards" which probably arose because Atmel reversed the pin order of PORTA:
Sanguino: PB0-7 0-7, PD0-7 8-15, PC0-7 16-23, PA0-7 31-24 <-- this last one is backwards 
Others: PB0-7 0-7, PD0-7 8-15, PC0-7 16-23, PA0-7 24 - 31 (all numbering ascending in port order)

I'd encourage you to include the first part of this patch to  OpenSprinkler.cpp  which initializes the PIN_EXP_SENSE to INPUT.

PR #27, opened 2015-09-23
Fixes for compilation.

ldav1s:
Added some exported variables to OpenSprinkler.h
Added const to PROGMEM variables.

I could not build without these changes using arduino 1.0.6 and Arduino-Makefile

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -49,11 +49,11 @@ ulong OpenSprinkler::checkwt_success_lasttime;
 char tmp_buffer[TMP_BUFFER_SIZE+1];       // scratch buffer
 
 #if defined(ARDUINO)
-  prog_char wtopts_filename[] PROGMEM = WEATHER_OPTS_FILENAME;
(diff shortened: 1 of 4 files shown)


ifreislich:
I think it's better to fix the definition of prog_char and prog_uchar.  I'm in the situation where these types are deprecated in my more recent AVR Tools (binutils-2.25.1, gcc-4.8.3, libc-1.8.1) and extern const ... doesn't work as expected hence pull-26.  Ray will probably say that compiling yourself is not supported, but at some point he's going to have to incorporate these changes to use a recent arduino and updated AVR Tools.

I do this in in defines.h:

diff --git a/defines.h b/defines.h
index 4e7782f..07850c8 100644
--- a/defines.h
+++ b/defines.h
@@ -221,6 +221,9 @@ typedef enum {
 
 #undef OS_HW_VERSION
 
+typedef const char prog_char;
+typedef const char prog_uchar;
+
 /** Hardware defines */
 #if defined(ARDUINO)

PR #28, opened 2016-01-17
Add GPIO Station support

PeteBa:
I modified the Special Station functionality that supports RF and IP stations to also control a raw GPIO pin. In essence this allows you to bypass the shift register/triac/protection circuitry of the OSPi and directly control a spare RPi gpio pin and any connected homebrew relay hardware. The /cs api command allows the setting of GPIO station type, gpio pin and active low/high.

I have a pull request for OS-App as well that provides the UI (see attached)  to associate any gpio pin with a station and set whether the hardware is active high or low.

Two areas that need some consideration:
1) GPIO Pin Contention - There in no validation of gpio usage in the Firmware but the UI restricts selection to avoid OSPi dedicated pins (i.e. shift register, RF, etc).
2) RPi R2 Only - The App assumes an OSPi with R2 connector. I couldnt see a way to detect if Arduino or RPi R1 to provide a tailored gpio selector.

Appreciate any feedback

gpio stations

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -882,6 +882,9 @@ void OpenSprinkler::switch_special_station(byte sid, byte value) {
     } else if(stn->type==STN_TYPE_REMOTE) {
       // request remote station
       switch_remotestation(stn->data, value);
+    } else if(stn->type==STN_TYPE_GPIO) {
+      // set GPIO pin
+      switch_gpiostation(stn->data, value);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


PeteBa:
Addressed the two areas of consideration above:
1) Added in validation of the Special Station data passed in the /cs api. This checks that the GPIO pin does not clash with an already assigned OSPi pin (e.g. shift register, flow meter, ...).
2) Added checking of the hardware version so that the selection of GPIO Station settings is restricted to OSPi only.

xaque208:
Oh goodie!

therevoman:
This is exciting!

PeteBa:
Yeah, it has been useful for me to both extend my OSPi in the back garden with a few extra relays and also to create a complete homebrew solution with RPiZero and WiFi Shield to link wirelessly in a station in the front garden. Curious what other will do with this as it opens up a lot of different control scenarios. Cheers, Pete

PR #29, opened 2016-02-21
Avoid port conflicts and allow hosting of Firmware, App and Weather on same local RPi

PeteBa:
Ray/Samer,

A joint pull-request to allow for Firmware, App and Weather servers to co-exist on the same local RPi2 box and avoid port conflict i.e. port 80 is used for both weather calls and UI page requests.

In order to do this, I needed to make a few small changes to server.js in Weather and weather.cpp in Firmware as follows:

1. Modified weather.cpp to parse both host and port from the specified weather url, i.e. allowing “localhost:3000” to be entered in the /su dialog. 
2. Modified server.js to allow both HOST and PORT to be specified in a .env file and set the precedence to be 1) environment variable, 2) .env file, 3) hardcoded default. This also allows HOST to be set to 0.0.0.0 giving access to the service from more than just localhost.
3. [Optional] Added a few additional pieces to help with diagnosing the set-up:
a.	Modified server.js to publish a service statement, “OpenSprinkler Weather Service”, when hit on the default page “/”.
b.	Modified weather.cpp to provide more diagnostic information on the connected server.

As an example, point 3 was helpful in figuring out that Debian translates “hostname” to 127.0.1.1 but “localhost” goes to 127.0.0.1. I hadn’t seen that before.

I don’t believe that any of the above changes would modify the default behavior but allow for more customised configurations for the users.

### DIFF
--- a/weather.cpp
+++ b/weather.cpp
@@ -185,24 +185,37 @@ void peel_http_header() { // remove the HTTP header
 
 void GetWeather() {
   EthernetClient client;
-
+  uint16_t port = 80;
+  char * delim;
   struct hostent *server;
+  
   nvm_read_block(tmp_buffer, (void*)ADDR_NVM_WEATHERURL, MAX_WEATHERURL);
+
+  // Check to see if url specifies a port number to use
+  delim = strchr(tmp_buffer, ':');
+  if (delim != NULL) {
+        *delim = 0;
+        port = atoi(delim+1);
+  }
+
   server = gethostbyname(tmp_buffer);
   if (!server) {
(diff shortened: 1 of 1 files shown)


PeteBa:
Hey Ray, wondered if you were happy with the weather.cpp commit above. I realise that it might be nicer to split port into a separate field but that would mean creating a new ADDR_NVM field to hold the port number. Rather than going down that route, I have overloaded the existing one to hold both url and port i.e. localhost:3000. Let me know if a problem and I'll rework. Samer has taken onboard the changes to server.js. Cheers, Pete

PR #33, opened 2016-06-04
Added support for HTTP Stations

PeteBa:
There are changes across both Firmware and App to support this feature.

The HTTP Station is a simple extension of the existing handler routines in both App and Firmware. This is very similar to the Remote Station code and allows the user to send HTTP GET commands to a server by configuring the following settings { Server, Port, On command, Off command }.

I have used a simple comma separated string to hold the above four parameters when transferring special data between App and Firmware using the api calls. This is a little different to the other Special Stations that use fixed width parameters. I did this because I wanted to support server names, in addition to IP addresses, and the On/Off commands can be arbitrary length.

In Firmware, I have extended the SpecialStationData structure from 23 to 120 bytes, i.e. the scratch buffer size, so that we can store complex On/Off commands (e.g. with hashed passwords or user tokens). This results in stns.dat file around four times the original size but still only ~4k so hoping that this is acceptable.

In the App, the number of Special Station Types is getting quite significant and the parameters were resulting in a really tall Station Attribute dialogue box. So I have introduced a “tabbed” version of the dialogue with Basic/Advanced pages. This keeps things a little more tidy.

Given the change to the stns.dat file format, this feature should really come in on a major version upgrade as it will require the user to do an export/import of settings to avoid stns.dat file corruption.

There are some things that I thought might be worth adding if you think useful:

- Specify whether a GET or POST message is used
- Select between “send and refresh” or “single shot” approach to keeping HTTP Station is sync
- Strip out any initial “/” entered by the user for the On or Off command entered into the App

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -20,6 +20,9 @@
  * along with this program.  If not, see
  * <http://www.gnu.org/licenses/>.
  */
+#if not defined(ARDUINO)
+#include <netdb.h>
+#endif
 
 #include "OpenSprinkler.h"
 #include "gpio.h"
@@ -792,15 +795,15 @@ static ulong hex2ulong(byte *code, byte len) {
 }
 
 /** Parse RF code into on/off/timeing sections */
-uint16_t OpenSprinkler::parse_rfstation_code(byte *code, ulong* on, ulong *off) {
+uint16_t OpenSprinkler::parse_rfstation_code(RFStationData *data, ulong* on, ulong *off) {
   ulong v;
-  v = hex2ulong(code, 6);
+  v = hex2ulong(data->on, sizeof(RFStationData::on));
   if (!v) return 0;
   if (on) *on = v;
-  v = hex2ulong(code+6, 6);
+	v = hex2ulong(data->off, sizeof(RFStationData::off));
   if (!v) return 0;
   if (off) *off = v;
-  v = hex2ulong(code+12, 4);
+	v = hex2ulong(data->timing, sizeof(RFStationData::timing));
   if (!v) return 0;
   return v;
 }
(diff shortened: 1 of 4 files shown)


JackTheDipper:
I'm highly interested in this feature. How can one try HTTP support running latest opensprinkler master? (This option is greyed out in the GUI)

PeteBa:
Jack, the code is present in both the app and firmware master branches to support this feature. However, the App is coded to check that the Firmware is version 2.1.7 before enabling the feature. This is because the HTTP Station functionality changes the settings file format and I didnt want to corrupt the user's existing settings file. If you really want to try this out then you can enable this feature by manually editing the version number in defines.h to 2.1.7 and rebuilding the Firmware. BUT before you do so you should export your configuration to a back-up location so that you can import your original settings once you have the new build up and running. Let me know how you get on. Cheers, Pete

JackTheDipper:
Thank you, Pete. Losing my settings would not be a problem for me as I'm new to the opensprinkler world. Actually, having some kind of HTTP support instead of relying on GPIO pins would enable me to use opensprinkler at all. ;-)

My valves are controlled by openHAB (and physically not by my pi which runs opensprinkler). I'd love to get the UI/app and the irrigation logic by opensprinkler, keeping openHAB in logical control of my valves. As openHAB provides a neat REST interface, I'd imagine opensprinkler could control my valves indirectly by sending HTTP requests to my openHAB instance. This way, opensprinkler could be used with a variety of different hardware setups (homematic valves, knx actuators, etc.).
Your hint was very useful, so now I'm able to select HTTP as station type. I hope that HTTP station is meant the way I want to use it... ;-) .. Do you plan to support HTTP Authentication?

(Btw, the command line still shows "failed to open gpio value for writing" warnings, although no GPIO station is setup anymore, heh)

JackTheDipper:
Is further work on this feature planned? For instance, HTTP Authentication, SSL (maybe too demanding for some microcontrollers?)?

In any case, I'd opt-in as a tester. ;-)

Issue #50, opened 2017-07-15
Docker container for OpenSprinkler on RPi


JohnStrunk:
I've been running OSPi using docker for a while, and I'm interested in sending a PR to get the Dockerfile into the main repo.

I'd like to get the log dir and NVM files into a proper docker volume. Currently, I have to do some symlink ugliness in the docker image because OpenSprinkler insists on writing its data files into the same directory as the executable due to get_runtime_path() and get_filename_fullpath().

My current idea is to get rid of the get_filename_fullpath calls and just do a chdir(datadir) in main(). This would allow a datadir to be passed in on the command-line or by default, have datadir=get_runtime_path().

Thoughts?

kpfleming:
This sounds like a good idea, even for non-Docker usage.

xaque208:
I'm on beagle.

frob:
I would use this.

Cabalist:
I would also use this.   It would be great to see this merged.

xaque208:
I've since purchased the new esp8266 platform and haven't looked back.    I'm still not sure the utility of this, since you'll need to interface with the GPIO, and passing that through the container, while perhaps possible, might be fragile.

PR #51, opened 2017-07-23
Containerize OpenSprinkler for RPi 

JohnStrunk:
* Add Dockerfile & README
* Remove get_filename_fullpath(). Use current dir instead
* Add '-d datadir' command line option. Default is path of OpenSprinkler exe
* Closes issue #50


JohnStrunk:
I have updated the Dockerfile to use the alpine image from the resin repo. This will allow docker hub/cloud to automatically build the container images.
I currently have the autobuild of my containerize branch available if anyone wants to give it a try:
`$ docker pull johnstrunk/opensprinkler-rpi`

criadoperez:
It's a pity this never made it to the main repository. Any interest in giving it a second try? Thanks.

JohnStrunk:
This PR has been sitting here for 3.5 years and was never even acknowledged by @rayshobby. If he is at least willing to commit to reviewing, I'll take a look at bringing it up to date.

Are you interested Ray?

rayshobby:
Sorry but I don't know what's the purpose of adding this containerizer. At the moment installing OpenSprinkler firmware on any linux system is just 2 lines of command: git clone, and build. Why do we need the containerize? I don't mean to dismiss the work, but I just don't know enough about this PR to understand its benefits.

criadoperez:
Maybe running it in any linux machine it can be two lines of code, but running it in docker, as you can see in @JohnStrunk code, it's more complicated than that.
@rayshobby , even if you don't use it in docker, some people do,  so if it doesn't harm you in any way, and this will help some current users or even bring in new docker users., it's a win-win for everyone. Of course, this is your call, but please consider accepting the pull request if @JohnStrunk is willing to review the code.

JohnStrunk:
That's a fair question. Distributing/running as a container has a couple benefits:

### Distribution
The software can be distributed as a single, pre-built unit that includes known versions of all the user-level libraries. People need not build from source, they can just `docker run` and it'll download the requested version and go.

### Run-time
Since containers have not only the application, but also all of its user-level dependencies, it greatly reduces the incidences of "but it works on *my* machine" that are caused by differing library versions. It also means users don't have to install any pre-req libraries-- They get exactly what you build and test with.

The above run-time benefits also extend to build-time if the build itself is containerized.

------

I personally use [Hypriot](https://blog.hypriot.com/downloads/) on my RPi's. The main reason is that I can take the image, edit a single txt file on my desktop, and I get a fully functioning headless system on my network (due to their use of cloud-init). This minimal provisioning step makes the SD cards largely disposable, so when one fails :man_shrugging: . I never need to attach a keyboard or monitor.

You may be correct... there may not be enough in the above that makes containerization important/useful for this project. That's completely your call.
I'm willing to take a look at re-doing this PR (I'll admit I'm still running the version from this PR :unamused:). I just want to know whether you're interested enough that I should spend the time on it.

Thanks for the consideration.
-John

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+
+########################################
+## 1st stage builds OS for RPi
+FROM base as build
+
+RUN [ "cross-build-start" ]
+
+RUN apk --no-cache add \
+      bash \
```
jbaudoux, line 12: I don't see bash required

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+      g++
+COPY . /OpenSprinkler
+RUN cd /OpenSprinkler && \
+    ./build.sh -s ospi
+
```
jbaudoux, line 17:
```suggestion

RUN apk del .build-deps

```

jbaudoux (commented):
Quick review, no test

rayshobby:
Could you explain this part: "Remove get_filename_fullpath(). Use current dir instead". This is in the firmware because it ensures the configuration files and also log files are written to the folder where the firmware exists. I haven't tested what's the result of removing this. But can you explain why you need to remove this function?

JohnStrunk:
Container instances are ephemeral... Each time they are run, they start from a pristine image. This causes a problem when trying to persist the config info across runs (like if the machine gets rebooted or software upgraded). Writing to a directory in the container image means it'll get destroyed when the container exits.
The change was to specifically allow storing the persistent data in its own directory that can be mapped to a docker volume (a directory that will persist longer than the life of the container). Basically, I need to separate the data (mutable) from the executable (static).

The changes cause it to use the "current directory" for all file ops, and add a bit of startup code to allow setting of the current dir. This preserves the existing behavior by doing a `cd` to the executable's directory by default, but adding the flag so that I can switch to the docker volume directory when in a container. This seemed like the least invasive way of allowing the location that is used for data to be changed.

developer79433:
Came here because I decided to invest my Saturday into containerising opensprinkler, it currently being the only workload on my raspberry pis that isn't containerised. Looks like @JohnStrunk has already done the bulk of the work required here - thanks, John!
What's blocking merging this? I see that the branch is now out of date and there are merge conflicts; I'll have a go at refreshing it and getting this merge-able again.
It really would be nice to get this last "odd man out" sorted so it runs nicely in docker on my rpis.
At one stage I was running a k8s cluster on the rpis. If I decide to do that again, I'll definitely want this to be containerised. Can't be manually installing software on k8s computes using bespoke installation processes like build.sh .

developer79433:
There's been too much change since you did this work, John. Merging/rebasing was just too hard.
Instead I started again, reusing some of your commits unchanged - see #225 .
I changed get_filename_fullpath instead of deleting it, which yields a smaller diff and hopefully less pushback from @rayshobby ?
Unfortunately I then got distracted cleaning up large numbers of compilation warnings.

rayshobby:
PR 262 has been merged. Closing this thread now.

PR #57, opened 2017-09-25
Feature Notifier - Extending integration options and flow related notifications

PeteBa:
This PR (and the associated PR for App) addresses three objectives with OpenSprinkler:
- Extend notification integration to get real-time OS event/usage into InfluxDb
- Get water usage at the Station and Program level rather than across a combined run
- Provide monitoring of flow rate to allow triggering of alarms if too high (i.e. leak/cut pipe)

During the process of implementing the above, the following have also been included:
- Specifying endpoints for IFTTT triggers so that you can send some events to, say, a spreadsheet log, and other to a phone notification
- Adding richer data to the notification messages i.e. Station End event also contains start/finish times, water level used, water volume usage and average flow rate
- Adding a generic WebHook interface for notifications that other services could hook onto. This is only “outbound” notifications and not an api.
- Allowing a “name” to be assigned to the unit that is sent with the notification to aid identification in multi-OS environments

In order to implement the above, the PRs have made a number of significant changes to both Firmware and App that require review:
- Generalised the IFTTT notification handler and UI to support multiple destinations (IFTTT, InfluxDb, WebHook)
- Moved much of the scheduling code from main.cpp to program.cpp to allow pre- and post- process of station and program scheduling
- Added flow volume to be apportioned across each of the running stations (done in a uniform distribution)

I have attached a comment to each of the commits below that sets out the headlines changes and attempts to explain some of the change rationale. In addition, there are a few questions that I would very much appreciate some feedback on that I have attached in the following PR level comment.

Appreciate your feedback and happy to take any modifications on-board.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@ nvm.dat
 OpenSprinkler
 wtopts.txt
 stns.dat
+ifttt.txt
+ifkey.txt
+name.txt
+webhook.txt
+influx.txt
--- a/EtherCard.h
+++ b/EtherCard.h
@@ -334,6 +334,10 @@ public:
                           const char *additionalheaderline, const char *postval,
                           void (*callback)(uint8_t,uint16_t,uint16_t));
 
+    static void httpPostRam (const char *urlbuf, const char *hoststr,
+                          const char *additionalheaderline, const char *postval,
+                          void (*callback)(uint8_t,uint16_t,uint16_t));
+
     static void httpPostVar (const char *urlbuf, const char *hoststr,
                           const char *urlbuf_varpart, const char *postval,
                           void (*callback)(uint8_t,uint16_t,uint16_t));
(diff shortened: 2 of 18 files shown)


PeteBa:
Ray, oviously the commits above touch a number of files and areas of the firmare so very keen to get your views and to adapt/refine as required. Some specific areas that would be great to follow-up on are:
- InfluxDb Inclusion - Adding support for InfluxDb may be "too specific" an integration. Concious that this brings additional maintenance burden. Are you comfortable with this?
- Scheduling Refactor - Changing the scheduler is like heart surgery - you don't want to get it wrong 8) - and so totally understand if you want me to change/adapt this. In fact, I am not sure the current split between main.cpp and program.cpp is optimal. I will follow up on this specifically.
- Notifier API - I have created a new notifier interface with specific calls for each of the notification type and a structure to pass the parameters that provides type checking. Are you OK with this?
- Socket Timeouts - I have introduced blocking network calls with a timeout to ensure that we don't inadvertently delay main loop() if there is a network issue and/or overwrite previous outgoing messages. Are there downsides to this ?
- Hardware Support - I have this running on HW2.3 but Visual Studio build exceeds memory of HW2.2 and lower. Do you still provide earlier hardware compatibility?

Some backward compatibility breaks that I want to make you aware of:
- All new features are auto-detecting when it comes to enabling the UI but the approach to two IFTTT notifications has been changed so be good to get your feedback on whether appropriate as follows:
    - IFTTT_FLOW changed to send real-time events (sent every 30sec) to allow monitoring/alarms. IFTTT_STATION_RUN event extended to include flow rate and volume.
    - IFTTT_WEATHER_UPDATE split into IFTTT_WATERLEVEL_UPDATE and IFTTT_IP_UPDATE as they would probably go to different event handlers in IFTTT (i.e. log on water level but alert on IP change)

I can revert these if we decide backward compatibility is required.

PeteBa:
Following feedback from Samer on App changes, I have updated Firmware to reference Site Name as the controller identifier sent with event notifications. Additionally, to support App import/export of config, I have bumped minor version so that we can trigger the necessary settings updates.

PeteBa:
Closing as stale.

PR #104, opened 2019-09-25
Add systemd service if applicable for autostart

waywardone:
More recent versions of Raspbian use `systemd` to manage application starts.
* Add `OpenSprinkler.systemd` - a `systemd` service template for OpenSprinkler
* Update `build.sh` to install the `systemd` service if appropriate.

NOTE: The check for the directory `/run/systemd/system` may not be foolproof across distros but is a good enough check for the Raspberry Pi and Raspbian IMO.

### DIFF
--- a/OpenSprinkler.systemd
+++ b/OpenSprinkler.systemd
@@ -1,0 +1,10 @@
+[Unit]
+Description=OpenSprinkler
+
+[Service]
+ExecStart=__OpenSprinkler_Path__/OpenSprinkler
+Restart=always
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
--- a/build.sh
+++ b/build.sh
@@ -18,7 +18,7 @@ else
 	g++ -o OpenSprinkler -DOSPI main.cpp OpenSprinkler.cpp program.cpp server.cpp utils.cpp weather.cpp gpio.cpp etherport.cpp -lpthread
 fi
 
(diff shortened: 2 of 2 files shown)


kpfleming:
Thanks for doing this; I'm about to do it as well, although I want to have the software running as a non-root user which complicates things somewhat.

OpenSprinkler.systemd
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=OpenSprinkler
```
WillCodeForCats, line 2: I added the following on my system to avoid network errors at boot time:
```
After=network-online.target
Wants=network-online.target
```

WillCodeForCats:
I tried running as a non-root user at first (as user `ospi` group `ospi`) and it worked for the most part, except for the reboot function in the UI didn't work. I installed under /opt/ospi with the following unit file:

/etc/systemd/system/opensprinkler.service 
```
[Unit]
Description=OpenSprinkler Raspberry Pi
After=network-online.target
Wants=network-online.target
 
[Service]
Type=exec
#User=ospi
#Group=ospi
Restart=on-failure
ExecStart=/opt/ospi/OpenSprinkler
WorkingDirectory=/opt/ospi

[Install]
WantedBy=multi-user.target
```

I didn't look to see how the reboot function is working in code to see if it could be modified to work at non-root.

rayshobby:
Thanks for the proposal. This has been addressed by the recent PR #295

PR #108, opened 2019-10-12
Enhance MQTT notifications

PeteBa:
This PR builds on the work from #102 and credit for the first commit of the PR has got to go to **jbaudoux** for starting the ball rolling on MQTT. The content has only been replicated here to facilitate rebasing onto the latest version 2.1.9 of Firmware and moving the content to a new feature branch. The second commit extends the functionality to allow configuring the MQTT Broker (server:port) via the App.

@jbaudoux, I asked Ray to create a new feature branch for this work so others can collaborate. I intend to work next on extending support for Last Will & Testament and Keep Alive, as well as, Arduino support. Appreciate your thoughts and involvement.

### DIFF
--- a/mqtt.cpp
+++ b/mqtt.cpp
@@ -1,0 +1,231 @@
+/* OpenSprinkler Unified (AVR/RPI/BBB/LINUX/ESP8266) Firmware
+ * Copyright (C) 2015 by Ray Wang (ray@opensprinkler.com)
+ *
+ * OpenSprinkler library
+ * Feb 2015 @ OpenSprinkler.com
+ *
+ * This file is part of the OpenSprinkler library
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#if defined(ARDUINO)
+	#include <Arduino.h>
+	#include <UIPEthernet.h>
+	#include <PubSubClient.h>
+	#if MQTT_KEEPALIVE != 60
+		#error Set MQTT_KEEPALIVE to 60 in PubSubClient.h
+	#endif
+#elif defined(OSPI)
+	#include <mosquitto.h>
+	#include <time.h>
+#endif
+
+#include "OpenSprinkler.h"
+#include "mqtt.h"
+
+#define MQTT_MAX_HOST_LEN		50 // Note: App is set to max 50 chars for broker name
+#define MQTT_RECONNECT_DELAY	60 // Minumum of 60 seconds between reconnect attempts
+
+#if defined(ARDUINO)
+	#if defined(ESP8266)
+		WiFiClient wifiClient;
+	#endif
+	EthernetClient ethClient;
+    struct PubSubClient *mqtt_client = NULL;
+#else
+	#if defined(OSPI)
+		#define MQTT_KEEPALIVE 60
+		struct mosquitto *mqtt_client = NULL;
+
+		static void mqtt_connection_cb(struct mosquitto *mqtt_client, void *obj, int rc);
+		static void mqtt_disconnection_cb(struct mosquitto *mqtt_client, void *obj, int rc);
+		static void mqtt_log_cb(struct mosquitto *mqtt_client, void *obj, int level, const char *message);
+	#else	// Do nothing os OSBO and DEMO
+		void * mqtt_client = NULL;
+	#endif
+#endif
+
+extern OpenSprinkler os;
+
+char OSMqtt::_host[MQTT_MAX_HOST_LEN + 1];	// IP or host name of the broker
+int OSMqtt::_port = 1883;					// Port of the broker (default 1883)
+char OSMqtt::_id[16];						// Id to identify the client to the broker
+bool OSMqtt::_enabled = false;				// Flag indicating whether MQTT is enabled
+
+// Create an MQTT instance. Note: on OSPi this also starts the loop thread.
+void OSMqtt::start(void) {
+	DEBUG_LOGF("MQTT Start: ");
+#if defined(ARDUINO)
+	Client * client = NULL;
+	bool wired_mac = false;
+	byte mac[6];
+
+    if (mqtt_client) { delete mqtt_client; mqtt_client = 0; }
+
+	#if defined(ESP8266)
+		if (m_server) client = &ethClient;
+		else client = &wifiClient;
+		wired_mac = (m_server != NULL);
+	#else
+		client = &ethClient;
+		wired_mac = true;
+	#endif
+	mqtt_client = new PubSubClient(*client);
+	os.load_hardware_mac(mac, wired_mac);
+
+	sprintf(_id, "OS-%02X%02X%02X%02X%02X%02X",mac[0], mac[1], mac[2], mac[3], mac[4], mac[5]);
+	DEBUG_PRINTF("Client %s ", _id);
+
+	if (mqtt_client == NULL) {
+		DEBUG_PRINTF("Failed to initialise client\n");
+		return;
+	}
+#elif defined(OSPI)
+	int rc;
+
+	mosquitto_lib_init();
+
+	if (mqtt_client) { mosquitto_destroy(mqtt_client); mqtt_client = NULL; };
+
+	mqtt_client = mosquitto_new(NULL, true, NULL);
+	if (mqtt_client == NULL) {
+		DEBUG_PRINTF("Failed to initialise client\n");
+		return;
+	}
+
+	mosquitto_will_set(mqtt_client, "opensprinkler/availability", strlen("offline"), "offline", 0, true);
+	mosquitto_reconnect_delay_set(mqtt_client, 2, MQTT_RECONNECT_DELAY, true);
+	mosquitto_connect_callback_set(mqtt_client, mqtt_connection_cb);
+	mosquitto_disconnect_callback_set(mqtt_client, mqtt_disconnection_cb);
+	mosquitto_log_callback_set(mqtt_client, mqtt_log_cb);
+#endif
+	DEBUG_PRINTF("Success\n");
+}
+
+#if defined(OSPI)
+// Callback called following a connect attempt. If the connection was successful then an "availability" message is sent.
+static void mqtt_connection_cb(struct mosquitto *mqtt_client, void *obj, int rc) {
+	DEBUG_LOGF("MQTT Connnection: %s\n", mosquitto_strerror(rc));
+	if (rc == MOSQ_ERR_SUCCESS)
+		mosquitto_publish(mqtt_client, NULL, "opensprinkler/availability", strlen("online"), "online", 0, true);
+}
+
+static void mqtt_disconnection_cb(struct mosquitto *mqtt_client, void *obj, int rc) {
+	DEBUG_LOGF("MQTT Disconnection: %s\n", mosquitto_strerror(rc));
+}
+
+static void mqtt_log_cb(struct mosquitto *mqtt_client, void *obj, int level, const char *message){
+	if (level != MOSQ_LOG_DEBUG )
+	DEBUG_LOGF("MQTT Log: %s (%d)\n", message, level);
+}
+
+#endif
+
+bool OSMqtt::enabled(void) { return _enabled; }
+
+// Configure the broker host and port fields and enabled/disable the interface
+// param config pointer to a string containing the JSON configuration settings in the form of "{server:"host_name|IP address", port: 1883, enabled:"0|1"}"
+void OSMqtt::setup(void) {
+	String config = os.sopt_load(SOPT_MQTT_OPTS);
+	if (config.length() == 0) {
+		_enabled = false;
+		DEBUG_LOGF("MQTT Setup: Config (None) Disabled\n");
+	} else {
+		int enable = false;
+		sscanf(config.c_str(), "\"server\":\"%[^\"]\",\"port\":\%d,\"enable\":\%d", _host, &_port, &enable);
+		_enabled = (bool)enable;
+		DEBUG_LOGF("MQTT Setup: Config (%s:%d) %s\n", _host, _port, _enabled ? "Enabled" : "Disabled");
+	}
+
+	if (mqtt_client == NULL) return;
+
+#if defined(ARDUINO)
+	if (mqtt_client->connected()) {
+		mqtt_client->disconnect();
+	}
+
+	if (_enabled) {
+		mqtt_client->setServer(_host, _port);
+		if (mqtt_client->connect(_id, NULL, NULL, "opensprinkler/availability", 0, true, "offline")) {
+			mqtt_client->publish("opensprinkler/availability", "online", true);
+		} else {
+			DEBUG_LOGF("MQTT Setup: Connect Failed (%d)\n", mqtt_client->state());
+		}
+	}
+#elif defined(OSPI)
+	mosquitto_disconnect(mqtt_client);
+
+	if (_enabled) {
+		int rc = mosquitto_connect(mqtt_client, _host, _port, MQTT_KEEPALIVE);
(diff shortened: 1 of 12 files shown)


PeteBa:
Support for OSPi, OS2.3 and OS3.0 (only tested on esp dev board). This also adds keep-alive messages with the broker and an availability topic.

Requires specific libraries to be installed. For OSPi you need the latest mosquitto libraries via `sudo apt-get install libmosquitto-dev`. For OS, you need the latest PubSubClient by knolleary.

mqtt.cpp
```diff
@@ -0,0 +1,231 @@
...
+void OSMqtt::loop(void) {
+	static unsigned long last_reconnect_attempt = 0;
+	unsigned long now = millis();
+
+	if (mqtt_client == NULL || !_enabled || os.status.network_fails > 0) return;
+
+#if defined (ARDUINO)
+	static int last_state = 999;
+	if (!mqtt_client->connected() && (now - last_reconnect_attempt >= (unsigned long)MQTT_RECONNECT_DELAY*1000)) {
+		if (mqtt_client->connect(_id, NULL, NULL, "opensprinkler/availability", 0, true, "offline"))
```
jbaudoux: Note here the issue with PubSubClient that has a timeout of 20 seconds. So each MQTT_RECONNECT_DELAY, you could lock the program on this line for 20 seconds, meaning that it won't react to anything.
I strongly recommend a non blocking client like AsyncMqttClient

jbaudoux: That remark is of course important when using the wifi on esp8266 with weak signal. That's a real pain. For wired connection, you shouldn't loose your connection and PubSubClient is an easy acceptable multi arch/board solution.

PeteBa: Yeah, I agree. I have the async libraries working but the problem child is OS2.3.

Both PubSubClient and UIPEthernet introduce timeouts (DNS Lookup, TCP Connection and MQTT Sync). The timeouts can be de-tuned but UIPEthernet does a lot more than just support MQTT.

The issue is really down to support for keep-alive and LWT. I'm reluctant to give up on these as "availability" is an important feature for "set and forget" devices. I'm going to play around a bit more to see if I can find a happy medium. But suspect we will need different solutions for each platform (i.e async mosquitto for OSPi, async-mqtt for OS3.0 and something cut down for OS2,3).

@jbaudoux , Let me know if you have thoughts/insight into OS2.3 solutions.

jbaudoux:
I think indeed that it can be a further enhancement. FYI. Esp supports
async tcp. Core provides lwip driver for embedded wifi chip and soon there
will be support for wired Ethernet with chip w5100/w5500. Full async tcp
stack wifi and Ethernet is nearly there. PR on core is under review.

I have ospi only, I'll test your PR with it. I don't know the differences
with the OS versions.

mqtt.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+	int rc = mosquitto_loop(mqtt_client, 0 , 1);
+	if ((rc == MOSQ_ERR_NO_CONN || rc == MOSQ_ERR_CONN_LOST) && (now - last_reconnect_attempt >= (unsigned long)MQTT_RECONNECT_DELAY*1000)) {
+		rc = mosquitto_reconnect(mqtt_client);
+		DEBUG_LOGF("MQTT Loop: Reconnect Status %s\n", mosquitto_strerror(rc));
+		last_reconnect_attempt = millis();
+	}
+
+	if (last_rc != rc) {
+		DEBUG_LOGF("MQTT Loop: Staus %s\n", mosquitto_strerror(rc));
```
jbaudoux:
```suggestion
		DEBUG_LOGF("MQTT Loop: Status %s\n", mosquitto_strerror(rc));
```

jbaudoux:
@PeteBa I tried your PR with OSPI but I cannot connect to Firmware with App 2.1.4. What should I do to make it work without loosing my config?

mqtt.cpp
PeteBa: Thanks.

PeteBa:
@jbaudoux , you need to either host your own version of the App with the associated MQTT PR included or you need to "hardcode" Firmware with your local settings. I suggest the later if you just want to test thing out.

To hardcode your settings then you can change Line 142 of `mqtt.c` as suggested below (change MQTT_IP and MQTT_PORT with your MQTT server IP/Port settings ):
```
--- String config = os.sopt_load(SOPT_MQTT_OPTS);
+++ String config = "\"server\":\"MQTT_IP\",\"port\":MQTT_PORT,\"enable\":1";
```

I would strongly recommend **against** using you main OSPi system to test this PR as there is no guarantee that your config will not be corrupted and that the system will operate properly.

jbaudoux:
@PeteBa That's what I did for testing and connection to MQTT is successful. But I could not connect with the App to test launching a station. I'm currently running my PR and I don't have that connection issue. I don't know what could be wrong :(

PeteBa:
Do you have firmware debug enabled ? You can edit `build.sh` and add `-DENABLE_DEBUG` to the OSPi build line. That should send mqtt diagnostics to the OSPi terminal. Let me know what that shows.

Also, what error do you get in your browser when navigating to `http://OSPI_IP:8080` ? If you are using Chrome then could you use the Chrome menu to go to "More tools -> Developer tools" and see if there are any error messages in the console view ?

jbaudoux:
Yes I enabled debug. First thing I did ;-)
```
root@pisprinkler:/home/pi/OpenSprinkler-Firmware-PeteBa-mqtt # ./OpenSprinkler
/home/pi/OpenSprinkler-Firmware-PeteBa-mqtt/
network established.
19-11-24 15:42:24 - MQTT Start: Success
19-11-24 15:42:24 - MQTT Setup: Config (None) Disabled
19-11-24 15:42:24 - MQTT Setup: Config (localhost:1883) Enabled
19-11-24 15:42:24 - MQTT Connnection: No error.
19-11-24 15:42:24 - MQTT Loop: Staus No error.
can't resolve http station - weather.opensprinkler.com
19-11-24 15:42:24 - MQTT Publish: opensprinkler/system {"state":"started"}
```

Can't use the local server as it's not connected to Internet and I don't host the UI locally.

Maybe due to the added options in the previous release that I don't run yet.

I'll leave it like that for now. I don't want to break my production system as I'm not going back to the site for some months.

PeteBa:
@jbaudoux , just looking a little closer at the log you included. It shouldn't be possible for those two different `MQTT Setup` lines to occur together as they are in the different blocks of the if/else conditional:
```
19-11-24 15:42:24 - MQTT Setup: Config (None) Disabled
19-11-24 15:42:24 - MQTT Setup: Config (localhost:1883) Enabled
```
Any chance you can double-check that part of the code (i.e. where you hardcoded the config settings from above) ?

jbaudoux:
That's because I harcoded mqtt config after the setup and copied the log
line. Mqtt connection is successful, so issue must be elsewhere.
Looking at latest release note, some config data is now moved to file.
Maybe that's why it didn't worked.

PeteBa:
@jbaudoux , just wondering if you have had a chance to take another look at this ? To help narrow down any issue, could you try the current master branch to see if that works. If it does, then that would suggest a problem with the PR. If you have a spare Raspberry PI then would be good to try with a clean install to ensure there is no configuration files corrupt.

jbaudoux:
I have spare rpi but no spare ospi shield to test. So without the hardware,
I don't know if that can be of any help

PeteBa:
You shouldnt need the OSPi shield to do basic testing as the RPI can run Firmware on its own (you just wont have physical relays/sprinklers to drive) but the web interface, mqtt functionality and basic operation (station on/off should work fine). It makes for a good stand-alone test environment if you dont want to disturb your main OSPI unit. 

From your note above, it seems that the App (2.1.4) wasnt connecting when using this PR. It would be good to see what is going on here so that we can fix if anything has been introduced incorrectly.

Also, you might want to update to the latest App version 2.1.6 but I dont think anything new was introduced in the recent releases that would impact the App connecting with Firmware but worth ruling out.

mqtt.cpp
PeteBa: Looking to close out this comment and appreciate any feedback.

There are currently three "timeouts" that impact MQTT network traffic and can block the main loop whilst establishing an MQTT connection as follows:

- DNS Lookup - A 15 second timeout if the network is down or the DNS server is unavailable
- TCP Connect - A 25 second timeout if the MQTT server port is closed
- MQTT Sync - A 15 second timeout if the MQTT Broker service is not responding

Note that these are not additive but rather the first timeout will exit the connection attempt so the maximum block would be 25 seconds (numbers based on OS2.3 which is worst case). This is similar for other outbound traffic such as HTTP Stations and IFTTT Notifications. However, MQTT is more susceptible to blocking as the "Keep Alive" communication happens on a very regular basis e.g. typically acks are sent every 15-60 second. Meaning that with default settings the main loop could be blocked for 25 seconds every minute if say the MQTT server was having issues.

I would like to maintain keep-alives as they provide a mechanism to detect OS going offline and allow for automations that alert the user. Very useful in "set and forget" equipment. Unfortunately, there aren't async versions of the libraries available across all platforms and for those platforms that do have libraries available (OSPi and OS3.0 WiFi) then all of the networking code base would need to be re-factored to transition to async communication. So I'm proposing is to remain with the existing synchronous architecture but adjust the KeepAlive and BackOff periods to maintain a balance between responsiveness and ensuring minimal disruption to the main event loop.

A very conservation approach would be to extend the KeepAlive period to 60 minutes (i.e. the Broker will recognise that OS is offline after an hour). To ensure the firmware doesn't block the event loop every hour for around 25 seconds then we can set a back-off period of 24 hours i.e. following an MQTT connection failure then Firmware would wait for 24 hours before attempting to reconnect. This is very conservative but would have limited impact on the main sprinkler functionality. The downside is that if there is a patchy network signal then the user may end up with significant number of "blackout" periods with no MQTT notifications.

An more aggressive approach would be to re-code the various timeouts in the libraries, say to 6 seconds. This would impact more than just MQTT functionality but would mean the KeepAlive period could be say 5 minutes and the back-off period down around an hour. It would mean modifying the libraries as they don't have apis for these settings.

@rayshobby , I would appreciate thoughts on this to get the balance right.

mqtt.cpp
PeteBa: Fixed in latest PR.

PeteBa:
@rayshobby , Hey Ray, the latest PR represents the majority of the changes to the basic MQTT functionality. There are updates to the message syntax that I would like to make and also to add in things like multi-site support but these will come in separate PRs. I am conscious that this PR is already getting fairly large so will pause here to get some feedback.

If there are others monitoring this thread that want to try out the functionality - I suggest in a test environment 8) - then let me know and I'll be happy to provide guidance on set-up. I would appreciate the feedback opportunity.

mqtt.cpp
jbaudoux: I think this issue is mainly in case of wifi connection with esp8266. Using asyncTCP is for sure the best way to solve this. And once esp8266 core 3.0.0 will be released, you could normally drop UIPEthernet. But that means splitting code in arduino for esp8266 and non esp8266.

PipeDeveloper:
I don't tested yet but its working for OSPi? Just asking before update to this beta firmware.
My plan is to link OpenSprinkler with my Home Assistant server.

craigcarps:
Hi Felipe, you can make rest URL calls direct from home assistant to
interact with open sprinkler. I’ve been using this method for the past 18
months and it’s been solid.

PipeDeveloper:
@craigcarps Hello!, very interesting, can you please send me an example of your config.yaml?
contact me on my e-mail: felipe.rojasv@mayor.cl
or on discord: Zelecktor#7398

Thank you!

craigcarps:
Hi Felipe

My configuration.yaml is:

switch openprinkler:
  - platform: command_line
    switches:
      frontyard:
        command_on: '/usr/bin/curl -X GET "http://<IPADDRESS>/cm?pw=<HASHEDPASSWORD>&sid=0&en=1&t=600"'
        command_off: '/usr/bin/curl -X GET "http://<IPADDRESS>/cm?pw=<HASHEDPASSWORD>&sid=0&en=0"'
        command_state: '/usr/bin/curl -X GET "http://<IPADDRESS>/js?pw=<HASHEDPASSWORD>&sid=0&en=0"'
        friendly_name: Front Yard Dripper
        value_template: "{{ value_json.sn[0] == 1 }}"

      backyard:
        command_on: '/usr/bin/curl -X GET "http://<IPADDRESS>/cm?pw=<HASHEDPASSWORD>&sid=1&en=1&t=600"'
        command_off: '/usr/bin/curl -X GET "http://<IPADDRESS>/cm?pw=<HASHEDPASSWORD>&sid=1&en=0"'
        command_state: '/usr/bin/curl -X GET "http://<IPADDRESS>/js?pw=<HASHEDPASSWORD>&sid=1&en=0"'
        friendly_name: Back Yard Dripper
        value_template: "{{ value_json.sn[1] == 1 }}"

And then i have some customisation going on:
switch.frontyard:
  icon: mdi:water-pump
switch.backyard:
  icon: mdi:water-pump
  device_class: opening
  initial_state: false

The trick is you'll need to hit up the open sprinkler API to get the instructions on doing a hash of your password, i cant recall. The API doco is great.

https://openthings.freshdesk.com/support/solutions/articles/5000716363

This switch that i've setup automatically triggers the zone for 10 minutes watering (600 seconds), i haven't used an variable input to pass to the switch.

Good luck
Craig

PipeDeveloper:
@craigcarps Hello. Im having a little issues, commands works at my browser also on my HA.
The issue is the state, it seems that HA reads the `command_state: '/usr/bin/curl -X GET "http:///js?pw=` it doesnt consider `&sid=0&en=0`, actually is a json format:

{"sn":[0,0,0,0,0,0,0,0],"nstations":8}

Here is the case: when i turn on a valve, trogling the icon on my HA, it returns off, but when i check on the opensprikler web UI, it turns on the valve. it seems that `value_template:` is wrong but if it is the case, HA is not recognizing correctly the state.

I really apreciate

craigcarps:
Hi Felipe

The sid value needs to correspond to the zone number, and also the value
template sn number needs to match the zone number too.

If the zone numbers don’t match then the feedback status will not work
correctly.

Craig

PeteBa:
I have hidden a few comments above as off-topic to de-clutter this PR. The comment discuss the REST api to OpenSprinkler and how it can be used to integrate with Home Assistant. They are still available to read for anyone interested.

PeteBa:
@PipeDeveloper , sorry for the delay responding as been on holiday. Yes, this works on OSPi but I suspect you may have found an alternative integration solution for HomeAssistant.

Note that this PR does not allow for controlling an OpenSprinkler unit but rather generating notifications when its activity changes. I have included instructions below if anyone wants to try out the PR. I suggest using a new SD card with the latest Raspbian Buster so that you can go back to your original setup when finished trying.

Firstly, ensure you have the right dependencies:
```
pi@raspberrypi:~ $ sudo apt-get update
pi@raspberrypi:~ $ sudo apt-get install git
pi@raspberrypi:~ $ sudo apt-get install libmosquitto-dev
```

Next, clone the OpenSprinkler repository and download the PR branch:
```
pi@raspberrypi:~ $ git clone https://github.com/OpenSprinkler/OpenSprinkler-Firmware.git
pi@raspberrypi:~ $ cd OpenSprinkler-Firmware/
pi@raspberrypi:~/OpenSprinkler-Firmware $ git fetch origin pull/108/head:featuremqtt
pi@raspberrypi:~/OpenSprinkler-Firmware $ git checkout featuremqtt
```

The PR is meant to be used with an equivalent PR on the App so that you can configure your MQTT server IP and Port address via the UI but if you just want to try this out then you can change **line 567 in main.cpp** from ```os.mqtt.begin();``` to ```os.mqtt.begin("server_ip", server_port, 1);``` where you replace server_ip with the ip address of your MQTT Broker and server_port with the port number (usually 1883). Note that the "1" is needed to set mqtt active.

Then build and run:
```
sudo ./build.sh
sudo .OpenSprinkler
```

You should see MQTT messages arrive in your broker when you turn stations on/off. let me know if you have any issues.

xaque208:
This is interesting, and looks reasonable.  I'm thinking of other ideas to build on top of, like reporting the wifi signal strength that I've seen in some other ESP8266 firmware by Anavi.

Nick-Adams-AU:
Is there anything outstanding before this can be merged into master? Does it need more testing?

PeteBa:
@Nick-Adams-AU , status is in the comments above.

Closing as I've moved on but code available for anyone to pick up and move forward.

ruimarinho:
@rayshobby this is so close and something we’ve been wanting for years! Any chance you can take from here?

PipeDeveloper:
MQTT always is better than an API. Also for control switches like a pump, read sensors, etc.

jbaudoux:
Thanks PeteBa for the enhancements. I have reopened a new PR https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/113 to continue the work.

PR #130, opened 2020-06-06
Mqtt api

vsoltan:
Added: 

- ability to connect to authenticated mqtt brokers. UI component is available at (https://github.com/OpenSprinkler/OpenSprinkler-App/tree/mqtt-secure). Menu->Integrations->MQTT has a username and a password field. 

- (proof of concept) ability to run a manual program by publishing a command to the broker. The following has been tested using mosquitto: 

  - mosquitto_pub -h hostname -p port -u username -P password -m 'opensprinkler/mpgm' -t 'pid=programnumber&uwt=(0 or 1 use weather for watering percentage)'

- for turning stations on and off, mqtt publishing payloads now contain the number of the appropriate station for better readability (when subscribing to 'opensprinkler/#' for exampe).

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1307,7 +1307,7 @@ void push_message(int type, uint32_t lval, float fval, const char* sval) {
 
 			if (os.mqtt.enabled()) {
 				sprintf_P(topic, PSTR("opensprinkler/station/%d"), lval);
-				strcpy_P(payload, PSTR("{\"state\":1}"));
+				sprintf_P(payload, PSTR("{\"state %d\":1}"), lval);
 			}
 			break;
 
@@ -1316,9 +1316,9 @@ void push_message(int type, uint32_t lval, float fval, const char* sval) {
 			if (os.mqtt.enabled()) {
 				sprintf_P(topic, PSTR("opensprinkler/station/%d"), lval);
 				if (os.iopts[IOPT_SENSOR1_TYPE]==SENSOR_TYPE_FLOW) {
-					sprintf_P(payload, PSTR("{\"state\":0,\"duration\":%d,\"flow\":%d.%02d}"), (int)fval, (int)flow_last_gpm, (int)(flow_last_gpm*100)%100);
+					sprintf_P(payload, PSTR("{\"state %d\":0,\"duration\":%d,\"flow\":%d.%02d}"), lval, (int)fval, (int)flow_last_gpm, (int)(flow_last_gpm*100)%100);
 				} else {
-					sprintf_P(payload, PSTR("{\"state\":0,\"duration\":%d}"), (int)fval);
+					sprintf_P(payload, PSTR("{\"state %d\":0,\"duration\":%d}"), lval, (int)fval);
 				}
 			}
 			if (ifttt_enabled) {
--- a/main.cpp
+++ b/main.cpp
@@ -570,11 +570,6 @@ void do_loop()
 	}
 	os.mqtt.loop();
 
-	if (curr_time > last_subscribe) {
-		os.mqtt.subscribe();
-		last_subscribe = curr_time + 5;
-	}
-
 	// The main control loop runs once every second
 	if (curr_time != last_time) {
 #if defined(ENABLE_DEBUG)
--- a/main.cpp
+++ b/main.cpp
@@ -404,6 +404,7 @@ void do_loop()
 
 	static ulong last_time = 0;
 	static ulong last_minute = 0;
+	static ulong last_subscribe = 0;
 
 	byte bid, sid, s, pid, qid, bitvalue;
 	ProgramStruct prog;
@@ -568,6 +569,11 @@ void do_loop()
 		os.status.req_mqtt_restart = false;
 	}
 	os.mqtt.loop();
+
+	if (curr_time > last_subscribe) {
+		os.mqtt.subscribe();
+		last_subscribe = curr_time + 5;
+	}
 
 	// The main control loop runs once every second
 	if (curr_time != last_time) {
--- a/main.cpp
+++ b/main.cpp
@@ -404,7 +404,6 @@ void do_loop()
 
 	static ulong last_time = 0;
 	static ulong last_minute = 0;
-	static ulong last_subscribe = 0;
 
 	byte bid, sid, s, pid, qid, bitvalue;
 	ProgramStruct prog;
--- a/mqtt.cpp
+++ b/mqtt.cpp
@@ -150,7 +150,7 @@ void OSMqtt::begin( const char * host, int port, bool enabled, const char *usern
 	DEBUG_LOGF("MQTT Begin: Config (%s:%d,%s) %s\n", host, port, *username ? "Secure" : "Unsecure", enabled ? "Enabled" : "Disabled");
 
 	strncpy(_host, host, MQTT_MAX_HOST_LEN);
-	_host[MQTT_MAX_ID_LEN] = 0;
+	_host[MQTT_MAX_HOST_LEN] = 0;
 	_port = port;
 	_enabled = enabled;
 
@@ -398,7 +398,7 @@ int OSMqtt::_connect() {
 	// connection is registered before we attempt to send our first NOTIFY_REBOOT notification. 
 	usleep(10000);
 
-	return MQTT_ERROR;
+	return MQTT_SUCCESS;
 }
 
 int OSMqtt::_disconnect(void) {
--- a/mqtt.cpp
+++ b/mqtt.cpp
@@ -102,8 +102,7 @@ char OSMqtt::_password[MQTT_MAX_PASSWORD_LEN + 1] = {0};
 
 // external declarations and function prototypes
 char mqtt_buffer[MQTT_MAX_MESSAGE_LEN];
-bool from_mqtt = false;
-void server_manual_program();
+void server_manual_program(char * buff);
 
 // Initialise the client libraries and event handlers.
 void OSMqtt::init(void) {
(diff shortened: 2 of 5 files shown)


rayshobby:
I would avoid using the term 'secure', instead, just call it authenticated, because my understanding is that mqtt user name and password are not sent through secure channels, so it's just a way to authenticate.

main.cpp
```diff
@@ -1301,7 +1307,7 @@ void push_message(int type, uint32_t lval, float fval, const char* sval) {
 
 			if (os.mqtt.enabled()) {
 				sprintf_P(topic, PSTR("opensprinkler/station/%d"), lval);
-				strcpy_P(payload, PSTR("{\"state\":1}"));
+				sprintf_P(payload, PSTR("{\"state %d\":1}"), lval);
```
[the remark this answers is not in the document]
rayshobby: It's a good note that the full topic is available (I suppose that's the result of -v?) I agree that the station number should be included as a separate variable, like 'id', instead of being included in the state variable.

main.cpp
```diff
@@ -569,6 +570,11 @@ void do_loop()
 	}
 	os.mqtt.loop();
 
+	if (curr_time > last_subscribe) {
```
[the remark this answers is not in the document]
vsoltan: I initially tried to set up subscriptions in the init function but, if anything happened to the connection, then I would have to resubscribe at some point. That's why I made it a loop, I chose 5 seconds because it would reduce the number of times the controller would have to resubscribe. But the callback makes a lot more sense here. I'll test it out and see if there are any issues.

mqtt.cpp
```diff
@@ -119,33 +134,49 @@ void OSMqtt::begin(void) {
...
+void OSMqtt::begin( const char * host, int port, bool enabled, const char *username, const char *password) {
+	DEBUG_LOGF("MQTT Begin: Config (%s:%d,%s) %s\n", host, port, username ? "Secure" : "Unsecure", enabled ? "Enabled" : "Disabled");
 
 	strncpy(_host, host, MQTT_MAX_HOST_LEN);
 	_host[MQTT_MAX_ID_LEN] = 0;
 	_port = port;
 	_enabled = enabled;
 
+	if (username) {
+		#define AUTH
```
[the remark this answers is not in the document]
vsoltan: This was some flawed logic that has been ironed out :)

mqtt.cpp
```diff
@@ -202,6 +245,18 @@ void OSMqtt::loop(void) {
...
 
+static void on_message_cb(const char topic[], byte *payload, unsigned int length) {
+	if (!strcmp(topic, MQTT_MANUAL_PROG_TOPIC)) {
+		for (int i = 0; i < length; ++i) {
+			mqtt_buffer[i] = payload[i];
+		}
+		mqtt_buffer[length] = '\0';
+		from_mqtt = true;
+		server_manual_program();
+		from_mqtt = false;
```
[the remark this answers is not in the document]
vsoltan: the problem with modifying the function signature is that server_manual_program is also included in a URLHandler array. But yeah definitely, I'll look into this.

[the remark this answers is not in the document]
PeteBa: Yeah, I saw that. But the above deals with the problem by retaining the existing `(void)` signature for URLHandler and adding a new `(char * p)` signature for use from within mqtt.cpp.

server.cpp
```diff
@@ -694,19 +694,19 @@ void manual_start_program(byte, byte);
...
-	if (from_mqtt) {
-		p = mqtt_buffer;
-	} else {
-		#if defined(ESP8266)
-			if (!process_password()) return;
-			if (m_client)
-				p = get_buffer;
-		#else
+void server_manual_program(void);
+void server_manual_program(char * p);
```
PeteBa, line 698: The function prototype should be moved to an include file that is shared with `mqtt.cpp`

mqtt.cpp
```diff
@@ -387,7 +364,7 @@ int OSMqtt::_init(void) {
 
 	if (mqtt_client) { mosquitto_destroy(mqtt_client); mqtt_client = NULL; };
 
-	mqtt_client = mosquitto_new("OS", true, NULL);
+	mqtt_client = mosquitto_new(NULL, true, NULL);
```
PeteBa, line 367: Note that I changed this to `NULL` as mosquitto will generate a random Id thus allowing multiple OS to connect to the same mqtt server. If PR #131 is merged then we have the option of using `load_mac_address()` in the same we do for Arduino.

mqtt.cpp
```diff
@@ -366,16 +340,19 @@ static void _mqtt_log_cb(struct mosquitto *mqtt_client, void *obj, int level, co
...
 			}
 			mqtt_buffer[message->payloadlen] = 0;
-			from_mqtt = true;
-			server_manual_program();
-			from_mqtt = false;
+			DEBUG_LOGF("Calling Manual Program\n");
+			server_manual_program(mqtt_buffer);
+			DEBUG_LOGF("Done Manual Program\n");
 		}
+		DEBUG_LOGF("cb done\n");
```
PeteBa, line 355: Can remove debug prints.

main.cpp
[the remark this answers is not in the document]
jbaudoux: I agree with @PeteBa that it's useless to repeat in the payload an information that is already in the topic

mqtt.cpp
```diff
@@ -71,24 +71,38 @@
 #endif
 
 extern OpenSprinkler os;
+extern char tmp_buffer[];
 
 #define MQTT_DEFAULT_PORT		1883	// Default port for MQTT. Can be overwritten through App config
 #define MQTT_MAX_HOST_LEN		50		// Note: App is set to max 50 chars for broker name
 #define MQTT_MAX_ID_LEN			16		// MQTT Client Id to uniquely reference this unit
 #define MQTT_RECONNECT_DELAY	120		// Minumum of 60 seconds between reconnect attempts
+#define MQTT_MAX_USERNAME_LEN 	50
```
jbaudoux, line 80: I think 20 should be sufficient for username and password

mqtt.cpp
```diff
@@ -332,9 +396,9 @@ int OSMqtt::_connect(void) {
 
 	// Allow 10ms for the Broker's ack to be received. We need this on start-up so that the
 	// connection is registered before we attempt to send our first NOTIFY_REBOOT notification. 
-	usleep(10000); 
+	usleep(10000);
 
-	return MQTT_SUCCESS;
+	return MQTT_ERROR;
```
jbaudoux, line 401: I don't understand this change

mqtt.cpp
jbaudoux: Also worth preventing buffer overflow like I did here: https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/133/files#diff-0f7914a36a22f82b449c1e46dcb60821R138

mqtt.cpp
```diff
@@ -119,25 +133,33 @@ void OSMqtt::begin(void) {
...
 }
 
 // Start the MQTT service and connect to the MQTT broker.
-void OSMqtt::begin( const char * host, int port, bool enabled ) {
-	DEBUG_LOGF("MQTT Begin: Config (%s:%d) %s\n", host, port, enabled ? "Enabled" : "Disabled");
+void OSMqtt::begin( const char * host, int port, bool enabled, const char *username, const char *password) {
+	DEBUG_LOGF("MQTT Begin: Config (%s:%d,%s) %s\n", host, port, *username ? "Secure" : "Unsecure", enabled ? "Enabled" : "Disabled");
 
 	strncpy(_host, host, MQTT_MAX_HOST_LEN);
 	_host[MQTT_MAX_ID_LEN] = 0;
```
jbaudoux, line 153:
```suggestion
	_host[MQTT_MAX_HOST_LEN] = 0;
```

jbaudoux:
Regarding "mpgm", any identifier should be placed in the topic and not in the payload. The payload should only contain a value. This is a good practice regarding the "retain" feature of MQTT.

main.cpp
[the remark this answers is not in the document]
jbaudoux: Anyhow, in general, an MQTT payload should not contain identifiers, this should always go in the topic

mqtt.cpp
PeteBa: Down to me. Fixed. Thanks.

mqtt.cpp
PeteBa: Down to me. Fixed. Thanks.

jbaudoux:
Thanks for the work @vsoltan . Better have feature branches when addressing multiple multiple subjects ;-p I appreciate the work to send MQTT messages to OpenSprinkler
It should include other commands and not be limited to only station on/off. I would be nice to be able to also send water level, rain delay, water percentage, ...

@PeteBa @rayshobby Maybe we should a bit rework the published messages to append the state in topic and only have a value in the payload. This should allow to have incoming messages with set in the topic. This will be more close to the messaging in home assistant. Like:
Send: 
* opensprinkler/system/state STARTED
* opensprinkler/station/1/state 0|1
* opensprinkler/station/1/report {duration, flow}
* opensprinkler/sensor1/state 0|1
* opensprinkler/sensor2/state 0|1
* opensprinkler/raindelay/state 0|1
* opensprinkler/sensor/flow/report {count, volume}

Receive:
* opensprinkler/station/1/set 1
* opensprinkler/raindelay/set 1
* opensprinkler/sensor/waterlevel/state 30  // would allow other device to report back to os the water level and have it displayed
* opensprinkler/station/1/set_waterpercentage 80  // I don't know if this can be per station, it would be nice
* opensprinkler/system/set_waterpercentage 80  // Set global water percentage

rayshobby:
Thanks for the PR. This feature has been added to PR #288.

Issue #124, opened 2020-05-27
Device unique identifier 


travisghansen:
I’ve been working on the home assistant integration for opensprinkler and some sort of unique identifier available via the api would be great. I’m not sure if there’s some sort of serial number available in the hardware or even the mac address if nothing else could do.

Thanks for the consideration!

xaque208:
There is some machine ID that you can reference I've seen in some other ESP projects.

travisghansen:
Any chance we can get it exposed via the api(s)?

psmedley:
It should be possible, opengarage exposes the Mac address via the api

rayshobby:
This has been addressed in the upcoming firmware (not officially released yet), so I am closing this issue for now.

PR #131, opened 2020-06-07
Expose MAC address on rest api

PeteBa:
Hey Ray, hope you are OK with this small PR that exposes MAC address via the rest api.

I have been working with the guys over at Home Assistant on an OpenSprinkler integration. It would be fantastic to combine the info coming from OS with the automation possible in HA. Specifically for things like Leak Detection (i.e. notify if flow is reading when no station is on) and similarly with failed valves (i.e. no flow when station on). The MQTT stuff will help take this even further!

One of the things we really need to make all this work is a unique identifier that stays pretty constant and that we can use to uniquely identify a particular OS unit. The MAC address is the most common identifier used for that purpose. Let me know if any concerns and I can modify as needed.

Addresses issue #124

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -574,6 +574,8 @@ void OpenSprinkler::reboot_dev(uint8_t cause) {
 #include "etherport.h"
 #include <sys/reboot.h>
 #include <stdlib.h>
+#include <sys/ioctl.h>
+#include <net/if.h> 
 #include "utils.h"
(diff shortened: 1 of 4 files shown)


rayshobby:
Sure, this is a good idea. I will take a look shortly.

psmedley:
IMHO it should be returned in a similar fashion to the same functionality on the open garage

PeteBa:
Addressed review comments

PR #113, opened 2020-03-22
Add MQTT notifications

jbaudoux:
Send MQTT messages. This implements this feature https://github.com/OpenSprinkler/OpenSprinkler-Firmware/issues/103
Events can be subscribed (like IFTTT).
Note that this PR does not allow for controlling an OpenSprinkler unit but rather generating notifications when its activity changes. 

Supported platforms: OSPI, OS3.0, OS3.2
On OSPI, it requires libmosquitto.

I also rewrite some IFTTT string message generation to use the standard sprintf instead of non-standard itoa.

This replaces my initial PR: https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/102
This replaces enhancement PR done by PeteBa: https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/108 Thanks to PetaBa for the review and improvements as well as the implementation for OS3

# OSPI instructions
If anyone wants to try out the PR. I suggest using a new SD card with the latest Raspbian Buster so that you can go back to your original setup when finished trying.

Firstly, ensure you have the right dependencies:
```
pi@raspberrypi:~ $ sudo apt-get update
pi@raspberrypi:~ $ sudo apt-get install git
pi@raspberrypi:~ $ sudo apt-get install libmosquitto-dev
```
Next, clone the OpenSprinkler repository and download the PR branch:
```
pi@raspberrypi:~ $ git clone -b ospi-mqtt https://github.com/jbaudoux/OpenSprinkler-Firmware.git
pi@raspberrypi:~ $ cd OpenSprinkler-Firmware/
```
The PR is meant to be used with an equivalent PR on the App so that you can configure your MQTT server IP and Port address via the UI but if you just want to try this out then you can change line 567 in main.cpp from os.mqtt.begin(); to os.mqtt.begin("server_ip", server_port, 1); where you replace server_ip with the ip address of your MQTT Broker and server_port with the port number (usually 1883). Note that the "1" is needed to set mqtt active.

Then build and run:
```
sudo ./build.sh
sudo .OpenSprinkler
```
You should see MQTT messages arrive in your broker when you turn stations on/off. let me know if you have any issues.

# OS3.x limitations
[...]

PR #132, opened 2020-06-08
MQTT: Fix keep alive

jbaudoux:
Relates to this comment: https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/113#issuecomment-640607304

### DIFF
--- a/mqtt.cpp
+++ b/mqtt.cpp
@@ -21,8 +21,6 @@
  * <http://www.gnu.org/licenses/>.
  */
 
-#define MQTT_KEEPALIVE 60  // Must be defined before including PubSubClient
-
 #if defined(ARDUINO)
 	#include <Arduino.h>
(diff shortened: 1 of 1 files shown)


PeteBa:
Setting `MQTT_KEEPALIVE` in context of `mqtt.cpp` does not change it in the context of `PubSubclient.cpp`.This is because when `PubSubClient.cpp` is compiled it only includes `PubSubClient.h` where MQTT_KEEPALIVE is set to default value.

If we want to alter this value - and I believe we should as every 60 seconds could impact the main loop - then we need to change the value in `PubSubClient.h` file and then test that it is what we expect.

jbaudoux:
@PeteBa Obviously. Thanks. Fixed now.

PeteBa:
Great stuff. Looks good to me.

PR #133, opened 2020-06-08
MQTT: Allow to connect with username and password

jbaudoux:
This follows up the several requests in https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/113 to support connection to broker with username and password.

Not tested.


PeteBa:
@jbaudoux see PR #130  for latest branch that includes username/password support.

rayshobby:
I've addressed the comments of this PR in #136 (including fixing a compilation error and minor changes to mqtt config to match the corresponding UI changes). Let me know if you have any comments/feedback. Thanks!

Issue #103, opened 2019-09-11
Feature: MQTT notifications


jbaudoux:
Allow to connect to MQTT broker.

Send notifications of the events:

- system start
  topic: opensprinkler/system
  message: {"state":"started"}
- station start
  topic: opensprinkler/station/<id starting at 0>
  message: {"state":1}
- station stop
  topic: opensprinkler/station/<id starting at 0>
  message:
    if flow sensor configured: {"state":0,"duration":[integer],"flow":[float]}
    else: {"state":0,"duration":<integer>}
- measured flow
  topic: opensprinkler/sensor/flow
  message: {"count":[integer],"volume":[float]}
- rain detection
  topic: opensprinkler/sensor/rain
  message: {"state":0/1}
- weather update
  not implemented yet

Allow to configure the events to be notified.

ruimarinho:
@jbaudoux this is great! Are you looking to add `/set` to switch stations on and off?

PeteBa:
@jbaudoux , some thoughts on the notifications.

- Multi-OS setups: I suggest extending the Topic string to include the Site name of the unit to differentiate multiple OS setups, e.g.  `opensprinkler/site/station/id`
- Flow Rate monitoring: having a regular notification, say every 30 seconds, with `volume` and `average flow rate` over that last period in the payload would let you monitor for leaking pipes and/or stuck valves
- Weather Call Failure: adding `errCode` to the payload of the Weather Call notification lets you track if there are problems getting an updated watering level and check if something has gone wrong 
- Alternative UI: I show the currently running watering cycle on the UI of my home automation system (Home Assistant). To do this I need to have `start`, `stop` times as well as `station name` and `water level` added in the Station Start payload so you can provide a real-time status display
- Station Run Stats: adding `duration`, `volume`, `average flow rate`, `water level` as payload fields in the Station Stop notification let me explore zone usage and water savings over time

I am sure there are a whole bunch of additional uses but these were the ones that I have made most use of with my set-up. Happy to provide a view on the message structure that I use.

nagyrobi:
Please add Water Level % and also Operation status (enabled/disabled as on/off) to MQTT notification.

StohanzlMart:
ADS 1115 with CUSTOM_SENSOR MQTT : https://github.com/StohanzlMart/OpenSprinkler-Firmware/commit/791d2be5bc61e12afe9359ed99b9946a22e60f24

hansonap:
[...]

PR #134, opened 2020-06-08
MQTT: refactor messages for better home assistant integration

jbaudoux:
I wonder if we should not change the way the messages are published to be more close to the several integrations existing in home assistant
https://www.home-assistant.io/integrations/mqtt/

This would allow in the future to also listen to "..../set" topics to receive commands

I know that json can be parsed in home assistant, but as the MQTT feature has not yet been released, it's still the time to ask our self if we should not do this small change to ease the integration. 

Relates to initial issue https://github.com/OpenSprinkler/OpenSprinkler-Firmware/issues/103 and my comment here https://github.com/OpenSprinkler/OpenSprinkler-Firmware/pull/130#issuecomment-640843714

Not tested

Before
--------
- system start
  topic: opensprinkler/system
  payload: {"state":"started"}
- station start
  topic: opensprinkler/station/<id starting at 0>
  payload: {"state":1}
- station stop
  topic: opensprinkler/station/<id starting at 0>
  payload:
    if flow sensor configured: {"state":0,"duration":[integer],"flow":[float]}
    else: {"state":0,"duration":<integer>}
- measured flow
  topic: opensprinkler/sensor/flow
  payload: {"count":[integer],"volume":[float]}
- rain detection
  topic: opensprinkler/sensor/rain
  payload: {"state":0/1}

After
------
- system start
  topic: opensprinkler/system/**power_state**
  payload: **on**
- station start
  topic: opensprinkler/station/<id starting at 0>/state
  payload: **{"state": "on"}** 
- station stop
  topic: opensprinkler/station/<id starting at 0>/state
  payload: 
    if flow sensor configured: **{"state": "off", "duration":[integer],"flow":[float]}**
    else: **{"state": "off", "duration":[integer]}**
- measured flow
  topic: opensprinkler/sensor/flow/state
  payload: {"count":[integer],"volume":[float]}
- rain detection
  topic: opensprinkler/sensor/rain/state
  payload: **"on"|"off"**

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1301,7 +1301,7 @@ void push_message(int type, uint32_t lval, float fval, const char* sval) {
 
 			if (os.mqtt.enabled()) {
 				sprintf_P(topic, PSTR("opensprinkler/station/%d/state"), lval);
-				itoa(1, payload, 10);
+				strcpy_P(payload, PSTR("{\"state\":\"on\"}"));
 			}
 			break;
 
@@ -1309,15 +1309,10 @@ void push_message(int type, uint32_t lval, float fval, const char* sval) {
 
 			if (os.mqtt.enabled()) {
 				sprintf_P(topic, PSTR("opensprinkler/station/%d/state"), lval);
-				itoa(0, payload, 10);
-				os.mqtt.publish(topic, payload);
-				topic[0] = 0;
-				payload[0] = 0;
-				sprintf_P(topic, PSTR("opensprinkler/station/%d/report"), lval);
 				if (os.iopts[IOPT_SENSOR1_TYPE]==SENSOR_TYPE_FLOW) {
-					sprintf_P(payload, PSTR("{\"duration\":%d,\"flow\":%d.%02d}"), (int)fval, (int)flow_last_gpm, (int)(flow_last_gpm*100)%100);
+					sprintf_P(payload, PSTR("{\"state\":\"off\",\"duration\":%d,\"flow\":%d.%02d}"), (int)fval, (int)flow_last_gpm, (int)(flow_last_gpm*100)%100);
 				} else {
-					sprintf_P(payload, PSTR("{\"duration\":%d}"), (int)fval);
+					sprintf_P(payload, PSTR("{\"state\":\"off\",\"duration\":%d}"), (int)fval);
 				}
 			}
 			if (ifttt_enabled) {
@@ -1374,7 +1369,7 @@ void push_message(int type, uint32_t lval, float fval, const char* sval) {
 
 			if (os.mqtt.enabled()) {
 				strcpy_P(topic, PSTR("opensprinkler/raindelay/state"));
-				itoa((int)fval, payload, 10);
+				strcpy_P(payload, (int)fval ? PSTR("on") : PSTR("off"));
 			}
 			if (ifttt_enabled) {
 				strcat_P(postval, PSTR("Rain delay "));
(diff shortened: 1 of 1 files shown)


jbaudoux:
cc @PeteBa @rayshobby @xaque208

xaque208:
Oh rad.  I'll upgrade for this change.

rayshobby:
I am not sufficiently familiar with Home Assistant to tell. From the firmware point of view, it would make sense to maximally reuse what the HTTP API already has. For example, to trigger a manual program run, the client can publish a message to topic /mp, with payload the same format as the HTTP API command (or to make it conform with json format, the payload could be something like {"cmd":"xxxxx"}. Reusing existing code paths can help minimize the amount of work and also make it easy to adapt when we introduce additional API commands in the future.

Nick-Adams-AU:
Should we also insert a unique id or MAC in the topic so we can support multiple units?

opensprinkler/00:11:22:33:44:55/system/power_state

jbaudoux:
The right approach for this would be to allow to configure the node name through the interface when MQTT is enabled. The node name is the first element in the topic and it is currently hardcoded to opensprinkler.

> Should we also insert a unique id or MAC in the topic so we can support multiple units?
> 
> opensprinkler/00:11:22:33:44:55/system/power_state

xaque208:
In terms of an ID for the device, I like this approach pretty well:

https://github.com/AnaviTechnology/anavi-thermometer-sw/blob/master/anavi-thermometer-sw/anavi-thermometer-sw.ino#L1722

You have to keep inventory either way, and I do still want to know the MAC somehow, but perhaps reading from `ESP.getChipId()` makes sense as well.

The rest of what you have there I think is a great start.

That Anavi code does something else interesting which is to publish a "config" about the object when a new MQTT client connects to the broker.  I've not dug in completely on how that is being achieved, but its easy to simulate with `mosquitto_sub`.  Perhaps the MAC could be part of that payload.

Nick-Adams-AU:
> In terms of an ID for the device, I like this approach pretty well

We might be covered already by the work done in https://github.com/OpenSprinkler/OpenSprinkler-Firmware/issues/124.

jbaudoux:
I have updated the PR with my proposal 2

nagyrobi:
Please push after boot not only availability, but also the state of the variables like current water level, rain delay, operation status.

PeteBa:
Some thoughts on the messages below:

> system start
topic: opensprinkler/system/power_state
payload: on

I think the `availability` topic handles the "powered on" scenario so we could slightly re-purpose this topic to notify of "system restart" given this can occur not just on a power cycle but also via watchdog timer and as soft-restart via the UI. Also, it would be useful to add the restart reason, i.e. based on `os.last_reboot_cause`,  to the payload to allow for an informative notification.

> station start
topic: opensprinkler/station/<id starting at 0>/state
payload: {"state": "on"}

Should we add intended duration to the payload given that it can vary based on watering level.

> station stop
topic: opensprinkler/station/<id starting at 0>/state
payload: if flow sensor configured: {"state": "off", "duration":[integer],"flow":[float]}
else: {"state": "off", "duration":[integer]}

If flow sensor is enabled then the flow value is only related to the last 30 seconds of the station run. I think it would be better to add `volume` and make `flow` the average during the run. Unfortunately, the flow and volume are not accurate if multiple stations are running at the same time but that is a limitation in the existing notifications as well.

> measured flow
topic: opensprinkler/sensor/flow/state
payload: {"count":[integer],"volume":[float]}

I know this matches IFTTT message but not sure of the use case for sending `count` given already sending `volume`.

This message is only generated when the programme queue is emptied and I'm not sure how people use this to be honest. I think we should move to something more like the App where the current flow rate is published say every 30 seconds. That would allow users to setup alarms if the flow rate is below a certain value when a particular station is running or flow is present when system is idle.

> rain detection
topic: opensprinkler/sensor1/state
payload: "on"|"off"

Should this use the more generic `opensprinkler/sensor/1/state` format similar to stations ?. I notice that you use 0|1 in the code rather than "on"|"off". Is there a reason for that ? Also, do we need a principle that all state information should be a json object e,g. `{ "state":"on", "type": "soil", ...}` to allow for additional information to be added in future ?

ruimarinho:
@jbaudoux do you plan on getting these changes merged? The MQTT version has been working great for me -- would be really interesting to start accept `/set` commands and getting more information vis-a-vis the REST API.

PR #136, opened 2020-06-09
MQTT: Allow to connect with username and password (replaces PR 133)

rayshobby:
This PR address the comments in PR #133


jbaudoux:
@rayshobby Thanks for the fixes. FYI, you should be able to push any fix on my branch instead of creating a new PR.

PR #138, opened 2020-06-13
MQTT support for multiple devices

PeteBa:
This adds support for multiple opensprinkler devices using MAC address as the unique client identifier. The topic tree is extended to the form `opensprinkler/OS-MACADDRESS/...` to allow for addressing specific units. I have retained `opensprinkler` as root topic to avoid namespace conflict and to allow for group broadcast in future.

There are some minor code corrections to allow for MQTT usernames without passwords.

### DIFF
--- a/mqtt.cpp
+++ b/mqtt.cpp
@@ -241,17 +241,16 @@ int OSMqtt::_init(void) {
 
 int OSMqtt::_connect(void) {
 	mqtt_client->setServer(_host, _port);
-	boolean state;
-	if (_username[0])
-		state = mqtt_client->connect(_id, _username, _password, MQTT_AVAILABILITY_TOPIC, 0, true, MQTT_OFFLINE_PAYLOAD);
-	else
-		state = mqtt_client->connect(_id, NULL, NULL, MQTT_AVAILABILITY_TOPIC, 0, true, MQTT_OFFLINE_PAYLOAD);
-	if (state) {
-		mqtt_client->publish(MQTT_AVAILABILITY_TOPIC, MQTT_ONLINE_PAYLOAD, true);
+
+	// Note: If (username == NULL) then password is ignored for anonymous login (i.e.no username or password)
+	//       If (username != NULL && password == NULL) then username is used without a password
+	if ( mqtt_client->connect(_id, _username[0] ? _username : NULL, _password[0] ? _password : NULL, availability_topic, 0, true, MQTT_OFFLINE_PAYLOAD) ) {
+		mqtt_client->publish(availability_topic, MQTT_ONLINE_PAYLOAD, true);
 	} else {
 		DEBUG_LOGF("MQTT Connect: Failed (%d)\n", mqtt_client->state());
 		return MQTT_ERROR;
 	}
+
 	return MQTT_SUCCESS;
 }
 
@@ -299,9 +298,9 @@ static bool _connected = false;
 static void _mqtt_connection_cb(struct mosquitto *mqtt_client, void *obj, int reason) {
 	DEBUG_LOGF("MQTT Connnection Callback: %s (%d)\n", mosquitto_strerror(reason), reason);
 
-	::_connected = true;
-
 	if (reason == 0) {
+		::_connected = true;
+
 		int rc = mosquitto_publish(mqtt_client, NULL, MQTT_AVAILABILITY_TOPIC, strlen(MQTT_ONLINE_PAYLOAD), MQTT_ONLINE_PAYLOAD, 0, true);
 		if (rc != MOSQ_ERR_SUCCESS) {
 			DEBUG_LOGF("MQTT Publish: Failed (%s)\n", mosquitto_strerror(rc));
@@ -345,13 +344,15 @@ int OSMqtt::_init(void) {
 
 int OSMqtt::_connect(void) {
 	int rc;
-	if (_username[0]) {
-		rc = mosquitto_username_pw_set(mqtt_client, _username, _password);
-		if (rc != MOSQ_ERR_SUCCESS) {
-			DEBUG_LOGF("MQTT Connect: Connection Failed (%s)\n", mosquitto_strerror(rc));
-			return MQTT_ERROR;
-		}
+	
+	// Note: If (username == NULL) then disable authentication
+	//       If (username != NULL && password == NULL) then username is used without a password
+	rc = mosquitto_username_pw_set(mqtt_client, _username[0] ? _username : NULL, _password[0] ? _password[0] : NULL);
+	if (rc != MOSQ_ERR_SUCCESS) {
+		DEBUG_LOGF("MQTT Connect: Connection Failed (%s)\n", mosquitto_strerror(rc));
+		return MQTT_ERROR;
 	}
+
 	rc = mosquitto_connect(mqtt_client, _host, _port, MQTT_KEEPALIVE);
 	if (rc != MOSQ_ERR_SUCCESS) {
 		DEBUG_LOGF("MQTT Connect: Connection Failed (%s)\n", mosquitto_strerror(rc));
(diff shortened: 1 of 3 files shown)


PeteBa:
@jbaudoux , appreciate a code review if you have the opportunity.

mqtt.cpp
```diff
@@ -329,29 +358,33 @@ int OSMqtt::_init(void) {
...
-	int rc;
-	if (_username[0]) {
-		rc = mosquitto_username_pw_set(mqtt_client, _username, _password);
-		if (rc != MOSQ_ERR_SUCCESS) {
-			DEBUG_LOGF("MQTT Connect: Connection Failed (%s)\n", mosquitto_strerror(rc));
-			return MQTT_ERROR;
-		}
+
+	// Note: If (username == NULL) then disable authentication
+	//       If (username != NULL && password == NULL) then username is used without a password
```
jbaudoux, line 381: No, you cannot do that. See https://github.com/eclipse/mosquitto/blob/master/lib/options.c#L69
Please revert this change

mqtt.cpp
```diff
@@ -93,16 +94,31 @@ char OSMqtt::_password[MQTT_MAX_PASSWORD_LEN + 1] = {0};	// password to connect
...
+/**************************** HELPERS ********************************************/
+
+int OSMqtt::_make_full_topic(char * full_topic, const char * topic)
+{
+#if defined(ENABLE_DEBUG)
+	if (strlen(MQTT_ROOT_TOPIC) + strlen(_id) + strlen(topic) + 2 > MQTT_MAX_TOPIC_LEN)
+		DEBUG_LOGF("MQTT Topic (%s) is too large. Will be truncated\n", topic);
+#endif
+
+	int len = snprintf(full_topic, MQTT_MAX_TOPIC_LEN, "%s/%s/%s", MQTT_ROOT_TOPIC, _id, topic);
```
jbaudoux, line 106: The root topic should be configurable through the app (like host, username, password). The MAC address should not part of the topic !!!

mqtt.cpp
PeteBa: Thanks for the review. This is getting interesting as it looks like the latest version of mosquitto does not match the documentation! But we should probably reference this link [here](https://github.com/eclipse/mosquitto/blob/30c818ed615ad078db3760fee1efd21d84bf5e64/lib/mosquitto.c#L229) as this is the version 1.3.4 which is available by `apt` for OSPi (i.e. much earlier). But lets dig into this as the code isnt operating as I would expect.

The motivation behind this change was to ensure we deal with three use cases: 1) username not provided (un-authenticated), 2) both username and password are provided (authenticated); and 3) username is provided but password is not (unprotected user login).

It is this third case that I was considering as according to the mosquitto documentation and code I linked above then we should be calling `mosquitto_username_pw_set(mqtt_client, _username, NULL)` rather than passing an empty string as _password.

Appreciate your thoughts on this as it feels like a decision between coding for the currently distributed version vs what might come down the road.

mqtt.cpp
PeteBa: So again thanks for the feedback and let me try to explain some of my thinking by taking this in two parts.

Firstly, the use of "opensprinkler" as the root topic. This is currently how the code exists today and available in the master branch. So this hasn't changed, I have just moved the definition out of `main.cpp` and into `mqtt.cpp` in order to better compartmentalise the code. I agree that in the future this should be configurable via the App and this PR moves us a little bit in that direction.

The second part is to address the situation where there are two or more opensprinkler devices on the network. This is the case for me and others. The current master branch is problematic as a user could enable mqtt on multiple devices with conflicting messages causing unexpected behaviour. So this is a first step to guarding against that with the information readily at hand. Again, a future PR could make this configurable from the App and the structure that this PR puts in place would make that a bit easier.

Note that the `load_hardware_mac()` routine produces a pretty stable identifier. Only recent OS3.0 have an option to change network interfaces and on OSPi I search the interface list in a particular order to minimise risk of change. In the future, I think leveraging the existing Site Name that App uses would be a good choice to differentiate opensprinkler devices as people are already familiar with it. Unfortunately, at the moment, it isn't accessible over the rest api and would require some changes to make that possible. I explored this some time ago [here](https://github.com/OpenSprinkler/OpenSprinkler-App/pull/42#issuecomment-333418131) and it looks to be a viable option.

Let me know your thoughts as I was keen to get something that makes the current master branch a bit safer while the bigger questions in #130 and #134 get worked through.

mqtt.cpp
jbaudoux: I had a wrong reading of source code. Apparently, it does not change anything but documentation says to implement it as you did.
https://mosquitto.org/api/files/mosquitto-h.html#mosquitto_username_pw_set
So, I agree with this change ;-)
Thanks

mqtt.cpp
jbaudoux: For enabling mqtt, you also have to provide a host. So let's show and propose to change the root topic as we did for username and password. This will allow multiple devices. And if you want to put your mac address in your topic, you are free to configure it like this.
But don't always append the root topic with the MAC address. For me, it is a bad design, and should not be the default behavior.
So default root topic:
"opensprinkler"
can be changed in the config to any other value like for example:
"opensprinkler1"
"opensprinkler/1"
"opensprinkler/00:0a:95:9d:68:16"

mqtt.cpp
rayshobby: I also pondered on this and so far I've assumed that if user name is non-empty then password is also non-empty. Would anyone intentionally set only a user name but empty password in practice?

mqtt.cpp
rayshobby: Two quite notes:
1. We do plan to add a 'device name', mainly to make it easy for users (with multiple OS) to tell which controller it's coming from when using, say IFTTT notifications (i.e. the device name will be included in the message). It's probably not appropriate to use 'device name' as mqtt root topic, because the device name can have all sorts of special characters and I am not sure if that's a good fit for root topic.
2. When I was working on integrating the mqtt branch, I did think about whether we should add an additional parameter - root topic - to give users the choice of deciding what topic to use. Related to my point 1 above -- what are the restrictions on the root topic name? For example, is it ok to have the 'space' character? What about other special characters?

mqtt.cpp
PeteBa: I agree, in normal practice you would want to use either case 1 or 2 but the spec does allow for case 3 as well. I can imagine scenarios in a trusted network where you want to track usage but aren't too worried about rogue agents ! 

Having said that, if we want to restrict against scenarion this then we should do it earlier in the process i.e. in the App UI or api response so we can provide user feedback.

mqtt.cpp
PeteBa: An mqtt topic is any utf-8, case-sensitive string. In many implementations the topic defaults to a slug of the device name or a readily available unique identifier (eg. mac).

Including a "device" leaf in the topic allows for multiple device support and having a "product" leaf in the topic allows for broadcast and discovery. So topics like "opensprinkler/device_id/endpoint" are quite common. However, there are alternative conventions and many sophisticated users will want to customise their own topics to their particular preference.

My intent with this PR is two fold: 1) safeguard multi-device owners so that  devices don't clash on the same topic and cause unpredictable behaviour, and 2) do it in a way that doesn't constrain the design of the message syntax under discussion in #134. Hence this PR doesn't expose the topic structure to end-user customisation.

mqtt.cpp
jbaudoux: As documentation says to do it like this and you implemented it, let's merge this improvement

mqtt.cpp
jbaudoux: > many sophisticated users will want to customise their own topics to their particular preference.

Exactly. So expose the root topic to end-user customization :-)

rayshobby: I am in favor of exposing the root topic to the user for customization. The topic can have an initial default value (it could either be 'opensprinkler' as we currently have, or it could be the device ID that @PeteBa proposed), and then user can change it to something else if they want.

rayshobby:
Thanks for the PR. This has been addressed by PR #288.

PR #145, opened 2020-07-06
Miscellaneous build fixes

emaste:
Attempting to compile on FreeBSD 13 with Clang and libc++ resulted in:

etherport.cpp:75:58: error: invalid operands to binary expression
    ('__bind<int &, sockaddr *, unsigned long>' and 'int')
        if (bind(m_sock, (struct sockaddr *) &sin, sizeof(sin)) < 0)
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -610,6 +610,7 @@ bool OpenSprinkler::load_hardware_mac(byte* mac, bool wired) {
 	mac[4] = 0x31;
 	mac[5] = iopts[IOPT_DEVICE_ID];
 
+#if !defined(__FreeBSD__)
 	if (m_server == NULL) return true;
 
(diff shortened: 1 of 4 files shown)


emaste:
First commit is a build fix when libc++ (rather than libstdc++) is used, now pushed a second commit to clean up/refactor build.sh somewhat, as part of my effort to build on FreeBSD.

rayshobby:
Thanks for the PR. I am closing it for now because there is currently no plan to support FreeBSD (nor do we intend to support it in the near future). Each additional platform adds testing overhead. We will continue to support Debian and Docker.

emaste:
Half of this isn't related to FreeBSD but is general maintainability / cleanup, would you like me to submit those in isolation or not bother?

PR #149, opened 2020-07-25
make.os23 file did not work on macOS Catalina

kraka40:
Only change worth noting here is the make.os23 file change with the error (?) in the identification of the PubSubClient Arduino library.  Would not compile without correcting the case of the library.

The server.cpp change was added to assist in debugging a situation I was having (precompile, from downloaded firmware) where password would be corrupted nearly daily.  I have not had this problem since compiling my own version and installing on OS 2.3 AC.

### DIFF
--- a/make.os23
+++ b/make.os23
@@ -1,12 +1,14 @@
 OFLAG = -Os
 ARDMK_DIR = .
-ARDUINO_DIR = $(HOME)/arduino-1.8.5
-ALTERNATE_CORE_PATH = $(HOME)/.arduino15/packages/MightyCore/hardware/avr/2.0.5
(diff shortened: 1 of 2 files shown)


rayshobby:
I am having trouble understanding this added line:
`bfill.emit_p(PSTR("***tmp_buffer:\"$S\""),tmp_buffer);`
this is going to append string in a non-standard way at the end of the json output. Is this intended? What's the purpose of this code?

rayshobby:
This has been addressed in the make file now:
https://github.com/OpenSprinkler/OpenSprinkler-Firmware/commits/master
the two alternative paths have been added in commented out section. I didn't change the library name of pubsubclient because the downloaded pubsubclient library as is it all small-case.

PR #150, opened 2020-08-04
Fixed refresh of remote stations so they go in order and code now sets correct duration on the remote initially and during refreshes.

cartridgebasic:
Modified the refresh routine to refresh stations in order so that all stations will be hit and be hit in a predictable order. Modified the calls involved to pass the station termination time when it's triggered so that the lower level routines can use the termination time to calculate duration based on current time and correct time can be set to the remote whether during the initial call or during later refreshes. Duration is stored in a variable that could be modified in the future by a pause routine so that duration is correct when things are unpaused.

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -37,6 +37,7 @@ byte OpenSprinkler::hw_rev;
 byte OpenSprinkler::nboards;
 byte OpenSprinkler::nstations;
 byte OpenSprinkler::station_bits[MAX_NUM_BOARDS];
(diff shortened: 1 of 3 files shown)


rayshobby:
A variant of the solution has been integrated into the code:
https://github.com/OpenSprinkler/OpenSprinkler-Firmware/commit/d8c1bc0987183c4cf7f598fad28944ec53400444
the added duration parameter for set_station_bit can't be integrated yet because of a conflict with the current plan for the upcoming firmware 2.2.0, but I've noted it in the todo list and will find a way to fit it in.

PR #158, opened 2020-12-31
Make firmware PlatformIO compilable

maxgerhardt:
Adds the needed `platformio.ini` plus a refactor of the `server.h` and `server.cpp` files because they confict with the Arduino-ESP8266 core files of `Server.h` and friends. This allows it to properly build on Windows which has a case-insenstive filenames and the wrong header file would be picked up, leading to compilation errors.

Folder structure is otherwise unchanged.

`platformio.ini` currently only builds the ESP8266 version of the fimware, but can also be expanded to build the AVR version of it.

Also see topic https://community.platformio.org/t/why-does-opensprinkler-firmware-for-esp8266-compile-in-visualmicro-and-arduino-but-not-in-this/18160

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -22,7 +22,7 @@
  */
 
 #include "OpenSprinkler.h"
-#include "server.h"
+#include "opensprinkler_server.h"
 #include "gpio.h"
 #include "testmode.h"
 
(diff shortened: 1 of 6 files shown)


rayshobby:
The changes from this PR have been integrated into the current master branch, including necessary changes to make the code compile for rpi. I didn't have platform.io installed so can't verify if it works with it now. But given the rather small number of changes, I assume it should be fine.

PR #171, opened 2021-04-22
Support retrieving group ids in scheduler-feature branch

rianadon:
I'm trying to finish the work of grouping stations in the `scheduler-feature` branch so that I can use the feature on my OpenSprinkler.

This PR is just a small fix to support retrieving the group ids from the api, so that the frontend can display this information. Of course the frontend is in a different repo, so I'd like to get the firmware code approved first before I submit the PR for the frontend. There, I can use the presence of the `stn_grp` attribute I'm adding to determine whether station grouping is supported.

### DIFF
--- a/OpenSprinkler.h
+++ b/OpenSprinkler.h
@@ -70,11 +70,14 @@ struct StationAttrib {	// station attributes
 	byte igs:1;	// ignore sensor 1
 	byte mas2:1;
 	byte dis:1;
+	byte unused:1;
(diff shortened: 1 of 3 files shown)


rianadon:
Also, the current setup only allows for 2 station groups (`gid` is stored in a single bit), which isn't very useful. Can I tag on a fix to use one of the 2 spare bytes in `StationAttrib` as the group id?

PR #173, opened 2021-04-26
Backwards-compatible version of scheduler-feature + more (grouped stations, date ranges, pause/resume)

rianadon:
This PR is a combination of @vsoltan's branch, combined with #171 which I just closed and a few changes on top.

I've edited the structures used in the [scheduler-feature](https://github.com/OpenSprinkler/OpenSprinkler-Firmware/tree/scheduler-feature) so that they maintain the same size, which means you could apply this PR, rebuild opensprinkler, and not lose any of your settings:
- The station struct using the 4-bit station group ids, which were already set aside!
- Using the final bit flag (previously unused) for programs to indicate whether the program should only run in a certain date range
- The program name max size is reduced from 32 to 28 to fit program start and end date

A few changes on top of @vsoltan's amazing work:
- The sequential flag is kept. With the flag set, the station uses its group id; otherwise, it's scheduled concurrently.
- The station group id is sent when returning a stations attributes so that the frontend can display it.
- Date ranges can be over new years (eg start Dec 1st, End Feb 1st).
- Dates are submitted to the server in numeric form, `(month << 5) + day`, rather than `4/26` form which makes parsing simpler. Plus it's easier to deal with only 1 form on the frontend.

@vsoltan's additions include scheduling programs with group ids, simplifying master station logic and adding support for more than 2, scheduling with date ranges, and pausing the queue.

I've only been working with grouped stations & date ranges, so that's been the most tested by me. There's probably a few more features of @vsoltan's in here that I haven't tested myself.

Most of the frontend work is done too, and I'll PR that soon once I add pause support. The latest version is [here](https://github.com/rianadon/OpenSprinkler-App).

# Testing instructions

If you'd like to test these changes, you'll first need to clone the fork. I recommend doing that in a separate directory so in case something goes wrong, it's easy to revert. Then copy all the `.dat` files holding your settings to the new directory and build.

```
git clone https://github.com/rianadon/OpenSprinkler-Firmware OpenSprinkler-Firmware-v2
cp OpenSprinkler-Firmware/*.dat OpenSprinkler-Firmware-v2/
cd OpenSprinkler-Firmware-v2
git checkout scheduler-extracompat
sudo ./build.sh ospi
sudo /etc/init.d/OpenSprinkler.sh restart
```

If you need to revert to the official OpenSprinkler software, run the build command and restart command in the original firmware folder (`cd OpenSprinkler-Firmware` and `sudo ./build.sh ospi` and `sudo /etc...`)

The changes for the frontend are not merged yet as well, so you'll need to change your OSPi's frontend url. Go to `http://<address-of-ospi>:8080/su`, and change "JavaScript URL" to https://rianadon.github.io/OS-App/js. Enter your password, and click "submit query". If successful, the frontend should say "unofficial build" at the top.


rayshobby:
I am checking this PR and will slowly integrate all the features into the firmware 2.2.0 branch. I started working on the date range feature without noticing this PR, and found I had similar thoughts (i.e. using the dummy bit for enable_daterange and move the parsing of the date to the UI instead of firmware :). I am going to keep the program name size the same, the reason being that firmware 2.2.0 is going to be compiled with ESP8266 core 3.0.2, with OTC (openthings) cloud connection enabled, and also it has changed to use LittleFS (because SPIFFS has a bug under core 3.0.2). Due to these changes, the flash memory will have to be wiped out anyways, so I am going to keep the program name size untouched (believe it or not, some users find even 32 characters too short...)

rayshobby:
Again, thanks for the PR. Most of the changes are now integrated into firmware 2.2.0 branch, with bug fixes as well. The UI changes are also integrated into the scheduler-ui branch with bug fixes. So I am closing this PR now.

rianadon:
That's awesome! I'm super excited to see these changes be officially released.
Thank you for all the hard work integrating them.

Issue #202, opened 2022-04-08
Unexpected event - "External IP updated"


Nu11u5:
For a few days now OpenSprinkler has been logging an event "External IP updated: xxx.xxx.xxx.xxx".  Since I have IFTTT webhooks set up I also get this as a notification.  In the 4 years I've had the OpenSprinkler set up I did not receive this type of event - now it is every few hours.  What's more, **the listed IP is never my public IP**, and for a while the event listed the "127.255.255.255" loopback address.

What is triggering this event?

What is used to determine the external IP?

CoupeWare:
I'm having the same problem.

salbahra:
Is the weather service working on the device?

I ask because I believe the weather service is what provides the external IP to the device.

CoupeWare:
Weather adjustment is set to manual. I just cleared the location since I'm not using the weather adjustment. Weather has not worked well since WU stopped working so I installed a rain sensor.

zhatt:
I am seeing this too.  It looks like the info coming back from the weather server is wrong.  Here an example of pulling the info via curl and eip keeps changing.

```
11:33 $ curl 'http://weather.opensprinkler.com/0?loc=42.36,-71.06&wto="key":""&fwv=219'
&tz=32&sunrise=342&sunset=1183&eip=2918528204&rawData={"wp":"Manual"}&errCode=0✔ ~/SRC/opensprinkler/git-OpenSprinkler-Firmware [master|✚ 4…3] 
11:33 $ curl 'http://weather.opensprinkler.com/0?loc=42.36,-71.06&wto="key":""&fwv=219'
&tz=32&sunrise=342&sunset=1183&eip=1822612243&rawData={"wp":"Manual"}&errCode=0✔ ~/SRC/opensprinkler/git-OpenSprinkler-Firmware [master|✚ 4…3] 
11:34 $ curl 'http://weather.opensprinkler.com/0?loc=42.36,-71.06&wto="key":""&fwv=219'
&tz=32&sunrise=342&sunset=1183&eip=1822612293&rawData={"wp":"Manual"}&errCode=0✔ ~/SRC/opensprinkler/git-OpenSprinkler-Firmware [master|✚ 4…3] 
11:34 $ curl 'http://weather.opensprinkler.com/0?loc=42.36,-71.06&wto="key":""&fwv=219'
[...]

PR #207, opened 2022-08-13
Fix issue #202 Unexpected event"External IP updated"

ssancese:
Ignore external IP address changes when notifying IFTTT
Minor changes to condition checking

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1014,14 +1014,12 @@ void do_loop()
 		// check weather
 		check_weather();
 
-		byte wuf = os.weather_update_flag;
-		if(wuf) {
-			if((wuf&WEATHER_UPDATE_EIP) | (wuf&WEATHER_UPDATE_WL)) {
-				// at the moment, we only send notification if water level or external IP changed
-				// the other changes, such as sunrise, sunset changes are ignored for notification
-				push_message(NOTIFY_WEATHER_UPDATE, (wuf&WEATHER_UPDATE_EIP)?os.nvdata.external_ip:0,
-																				 (wuf&WEATHER_UPDATE_WL)?os.iopts[IOPT_WATER_PERCENTAGE]:-1);
-			}
+		if( os.weather_update_flag & WEATHER_UPDATE_WL)
+		{
+			// at the moment, we only send notification if water level IP changed
+			// the other changes, such as sunrise, sunset changes are ignored for notification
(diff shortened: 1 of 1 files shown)


zhatt (commented):
This just disables the IP change reporting. The real fix that is needed is in handling of the proxy in the weather server https://github.com/OpenSprinkler/OpenSprinkler-Weather/blob/286e87c812f461f46f0f9e6fd04ff62aaf178384/routes/weather.ts#L172

ssancese:
@zhatt you might be right, I'm not that deep in the understanding of the whole thing but I just can't understand why IP change should be reported along with weather info.
IP changes should be treated  and reported separately and possibly reporting should be enabled by the user through UI settings.

rayshobby (approved):

rayshobby:
External IP (eip) change is useful for those who use port forwarding for remote access. The reason it is coupled with weather update is that eip change is sent back to the controller through weather query -- the controller itself cannot detect eip change, only an external server can, so while it seems unrelated to weather change, this is the easiest way to detect eip change. This is similar to time-zone change -- it isn't directly related to weather data but it's convenient to detect time-zone change in the same weather query. I agree that the real fix is to fix the weather server so it won't cause false positives about eip change. For now I will accept your PR. The upcoming firmware 2.2.0 supports OpenThing Cloud (OTC) connection, people won't need to set up port forwarding any more for remote access. Thus it's not longer necessary to get notified about eip change. In the future we will fix the weather server code and probably add a different flag for eip change.

Nu11u5:
I understand that there is a technical case for communicating an external IP change to the controller, but the feature seems to not be working correctly.

I am constantly getting EIP change notifications, and every time the listed IP has nothing to do with my ISP connection.

Usually what is returned is in the 162.158.62/23 network, which is owned by Cloudflare.  Sometimes it even replies with 127.0.0.1.  I suspect there is an issue with the EIP detection service, which I am guessing is hosted behind Cloudflare.

My router's WAN IP is public and not NATed, and I am not connecting through a VPN service.

rayshobby:
I am not sure if you read my message: I acknowledged that the EIP change has false positives, so it's sending notifications when it shouldn't. Therefore I merged this PR which disables EIP change notification for now. Given that firmware 2.2.0 no longer requires port forwarding, most users won't need to know their EIP anymore.

PR #221, opened 2023-01-26
Branch ospi - analog sensor api ospi compatibility

opensprinklershop:
OSPi does not support the analog sensor box, but it supports the network sensors or remote sensors
This branch allows compiling the new analog sensor api with the ospi


DomiStyle:
Since the OSPi has an PCF8591T ADC chip onboard, would it be possible to use the UI with those pins?

opensprinklershop:
Yes it's possible. Also I'm working on it, but currently I can't get it working.

DomiStyle:
Thanks, that would be exactly what I need. Looking forward to it.

Feel free to let me know when you need testers.

DomiStyle:
Thanks! Will give it a try soon.

rayshobby:
Features in this PR are included in the latest analog sensor PR.

PR #230, opened 2023-05-12
feat: prepare for OrangePi boards

paulkram:
I had a OrangePi PC board laying around and wanted to use it with OpenSprinkler. It's based on a Allwinner H3. According to the documentation, all OrangePi One/Lite/Pc/Plus/PcPlus/Plus2e should be usable with OpenSprinkler with my changes immediately.

Works with Armbian 23.02 Jammy

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -93,7 +93,7 @@ extern ProgramData pd;
 	LiquidCrystal OpenSprinkler::lcd;
 	extern SdFat sd;
 #else
-	#if defined(OSPI)
+	#if defined(OSPI) || defined(OSOPI)
 		byte OpenSprinkler::pin_sr_data = PIN_SR_DATA;
 	#endif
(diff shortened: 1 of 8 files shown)


N3szt0r:
I tried on Orange Pi Zero. Build.sh ifarray runs out to else. My $1 value is ospi. I dont know how. Installed OS is Armbian 23.02 jammy

paulkram:
Hi,
please try with $1 value opipc. This should do the trick. Please double check in wiringOp docu if the pin layout fits.

rayshobby:
Thanks for the PR request. I am closing it for now because we've made fairly significant changes to the OSPI branch, to support I2C LCD and buttons. Modifying it for Orange Pi would require a significant amount of work and testing. We currently do not support Orange Pi, nor do we intend to support it in the near future (each additional supported hardware entails significant amount of testing efforts, which we unfortunately can't afford). Thank you.

PR #238, opened 2023-06-28
Mqtt multi device

repa6:
Hello Ray!

This PR addresses changes in MQTT topic path. The changes help users with more controllers to distinguish them. Each device gets its own mac address into the MQTT topic path.

After modification the MQTT topic looks like:
328294096_3332586880335871_5334181142344244116_n

### DIFF
--- a/mqtt.cpp
+++ b/mqtt.cpp
@@ -48,21 +48,21 @@
 #if defined(ENABLE_DEBUG)
 	#if defined(ARDUINO)
 		#include "TimeLib.h"
-		#define DEBUG_PRINTF(msg, ...)    {Serial.printf(msg, ##__VA_ARGS__);}
-		#define DEBUG_TIMESTAMP(msg, ...) {time_t t = os.now_tz(); Serial.printf("%02d-%02d-%02d %02d:%02d:%02d - ", year(t), month(t), day(t), hour(t), minute(t), second(t));}
+		#define DEBUG_PRINTF(msg, ...)		{char buffer[TMP_BUFFER_SIZE]; sprintf(buffer, msg, ##__VA_ARGS__); Serial.print(buffer);}
+		#define DEBUG_TIMESTAMP(msg, ...)	{time_t t = os.now_tz(); char buffer[TMP_BUFFER_SIZE]; sprintf(buffer, "%02d-%02d-%02d %02d:%02d:%02d - ", year(t), month(t), day(t), hour(t), minute(t), second(t)); Serial.print(buffer);}
 	#else
 		#include <sys/time.h>
-		#define DEBUG_PRINTF(msg, ...)    {printf(msg, ##__VA_ARGS__);}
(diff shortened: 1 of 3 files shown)


mqtt.cpp
```diff
@@ -48,21 +48,21 @@
 #if defined(ENABLE_DEBUG)
 	#if defined(ARDUINO)
 		#include "TimeLib.h"
-		#define DEBUG_PRINTF(msg, ...)    {Serial.printf(msg, ##__VA_ARGS__);}
-		#define DEBUG_TIMESTAMP(msg, ...) {time_t t = os.now_tz(); Serial.printf("%02d-%02d-%02d %02d:%02d:%02d - ", year(t), month(t), day(t), hour(t), minute(t), second(t));}
+		#define DEBUG_PRINTF(msg, ...)		{char buffer[TMP_BUFFER_SIZE]; sprintf(buffer, msg, ##__VA_ARGS__); Serial.print(buffer);}
+		#define DEBUG_TIMESTAMP(msg, ...)	{time_t t = os.now_tz(); char buffer[TMP_BUFFER_SIZE]; sprintf(buffer, "%02d-%02d-%02d %02d:%02d:%02d - ", year(t), month(t), day(t), hour(t), minute(t), second(t)); Serial.print(buffer);}
```
Szekyy, line 52: Are years represented in two-digit format?

mqtt.cpp
```diff
@@ -71,28 +71,46 @@
...
-char OSMqtt::_password[MQTT_MAX_PASSWORD_LEN + 1] = {0};  // password to connect to the broker
-int OSMqtt::_port = MQTT_DEFAULT_PORT;  // Port of the broker (default 1883)
-bool OSMqtt::_enabled = false;          // Flag indicating whether MQTT is enabled
+#define MQTT_KEEPALIVE			60
+#define MQTT_DEFAULT_PORT		1883	// Default port for MQTT. Can be overwritten through App config
+#define MQTT_MAX_HOST_LEN		50		// Note: App is set to max 50 chars for broker name
+#define MQTT_MAX_USERNAME_LEN	32		// Note: App is set to max 32 chars for username
+#define MQTT_MAX_PASSWORD_LEN	32		// Note: App is set to max 32 chars for password
+#define MQTT_MAX_ID_LEN			16		// MQTT Client Id to uniquely reference this unit
+#define MQTT_MAX_TOPIC_LEN		60		// Includes ROOT_TOPIC, DEVICE_TOPIC and endpoint
```
Szekyy, line 80: Are 60 characters enough for every possibility? MQTT documentation states that the maximum length is 65535, 128 bytes would be more forgiving.

repa6:
@rayshobby , could you please take a look. Any suggestions are welcome to this topic. I have some friends suffering with multiple devices from the non-unique topic.

mqtt.cpp
repa6: these are main repo changes

mqtt.cpp
repa6: these are main repo changes

rayshobby:
Thanks for the PR. This has been addressed in PR #288. We used a slightly different method, which allows the user to specify a custom publishing topic. It can be the MQTT ID as you used, or it can be anything else.

PR #240, opened 2023-07-06
If PIN_RFTX == 255, disable all code that uses that pin.

jdchristensen:
This makes it safer to set PIN_RFTX to 255, so pin 15 can be used for other purposes.

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -865,9 +865,11 @@ void OpenSprinkler::begin() {
 	nboards = 1;
 	nstations = nboards*8;
 
-	// set rf data pin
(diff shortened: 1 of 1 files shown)


jdchristensen:
I plan to explain how one can use pin 15 to add an extra station, and having this merged makes the diff smaller while also making the code more robust.

rayshobby:
The change in begin() function isn't necessary because both pinModeExt and digitalWriteExt will return right away if the pin number is 255. But in any case, I'm merging it to the master branch.

jdchristensen:
Thanks, I wasn't sure what `pinModeExt` and `digitalWriteExt` would do if passed 255.

Issue #248, opened 2023-08-27
Flow Sensor not resetting for stations without flow meter


bb09961:
Only a few of my stations are plumbed in to my flow meter.   When a station without the flow meter runs, the previous station's flow rate is stored in the logs.  It should store a 0.     The attached screenshot shows the entries in red that should be 0. The values in green are correct.    This issue shows up in my case, but also would be present if a station did not water at all due to a valve not opening.

![Flow_Rate_Not_Resetting](https://github.com/OpenSprinkler/OpenSprinkler-Firmware/assets/143335221/b12449d2-6b5c-401b-bcfe-29a074755860)

Based on the comments in flow_poll() function, this IF statement jumps out of the function if the sensor never drops from HIGH to LOW,   OR it would also jump out if the sensor never changes at all.  This is the case for my stations without a flow meter.

```
void flow_poll() {  
...  
	if(!(prev_flow_state==HIGH && curr_flow_state==LOW)) { // only record on falling edge  
		prev_flow_state = curr_flow_state;  
		return;  
	}
``` 

If the sensor never changes, then it will never get to execute any of the following initialization code inside flow_poll() :
```
	/* RAH implementation of flow sensor */
	if (flow_start==0) { flow_gallons=0; flow_start=curr;} // if first pulse, record time
	if ((curr-flow_start)<90000) { flow_gallons=0; } // wait 90 seconds before recording flow_begin
	else {	if (flow_gallons==1)	{  flow_begin = curr;}}
	flow_stop = curr; // get time in ms for stop
	flow_gallons++;  // increment gallon count for each poll
	/* End of RAH implementation of flow sensor */
```

Would it work to move or copy the "flow_gallons=0"  from flow_poll()  to the turn_on_station() function where the flow_start variable is also reset?
```
void turn_on_station(byte sid, ulong duration) {
	// RAH implementation of flow sensor
	flow_start=0;
	//Move or copy flow_gallons reset here?//
	flow_gallons=0
```

If flow_gallons=0 is moved, then the existing code in the turn_off_station() function should work properly to log a 0 for flow_last_gpm:

```
	// RAH implementation of flow sensor
	if (flow_gallons > 1) {
		if(flow_stop <= flow_begin) flow_last_gpm = 0;
		else flow_last_gpm = (float) 60000 / (float)((flow_stop-flow_begin) / (flow_gallons - 1));
	}// RAH calculate GPM, 1 pulse per gallon
[...]

PR #252, opened 2023-09-25
Flow Alerts

bb09961:
### This code change adds optional Flow Alerts to the OpenSprinkler firmware with no required code changes to the OpenSprinkler UI.

Flow Alerts will be sent via Email, IFTTT or through an MQTT message after a station run is complete if the flow rate exceeds a station's set point.  

The station flow rate is calculated by pre-existing firmware logic that divides total volume by total duration in minutes after a station completes a run that is 2.5 minutes or longer. 

A station's flow rate set point must be stored as a suffix to the station's name.

**** Flow Rate Alerts are meant for **sequential stations ONLY** and will not work for stations running in parallel.
**** However, Flow Rate Alerts WILL work correctly with stations that are tied to a Master Station controlling a master valve or pump.
.
.
### Steps to Enable Flow Alerts:
- A flow meter should be configured on Sensor 1 in the OpenSprinkler UI with the correct Flow Pulse Rate factor.  
- Assign flow rate set points in gallons(or liters) per minute by appending the set point value to the END of each station's name. 
- Station set points must consist of a 5 character number including a decimal point with no trailing spaces. It can be any combination of digits, but must be a total of 5 characters. (For example: 15.10, 05.10, 5.100, 510.0 or 0.000) 
- If a set point is not found or not formatted correctly, flow alerts will be turned off for that station.
- Use the OpenSprinkler user manual to configure OpenSprinkler for Email notificiations,  IFTTT integration or configuring and enabling MQTT correctly.
- Flow Alerts are not linked to a specific notification event type in the OpenSprinkler UI.  Instead, Flow Alerts through Email, IFTTT or MQTT will be enabled for each station individually whenever:
-- Email notifications are enabled, an IFTTT key is found or MQTT is enabled
-- a flow meter is configured
-- a valid numerical set point is found at the END of the station name
### Testing Flow Alerts:
- Assign all station set points to 0.000 and run each station for 3-4 minutes to test the flow alerts and obtain the current flow rate for each station.
- Pre-existing OpenSprinkler logic requires a station to run for at least 2.5 minutes to calculate the flow rate.
- It is recommended to run stations for their full normally scheduled durations to determine the most accurate baseline flow rates. 
- Flow rates can vary somewhat with city water pressure or after very short run times, so determining an optimal set point may be an iterative process.

### Leak Detection:
- Flow Alerts can also be used for system leak detection:  
-- Configure an unused\non-wired station with a set point of 0.000
-- Run the station once a day for 1 hour.  If any flow is detected an alert will be sent.
-- This will detect any leaks between your flow meter and station valves or the valves themselves.

  

.
### Example of station name with Flow Alert Set Point suffix:
image

  

.
### Example of Flow Alert Email Notification:
example_email

  

.
### Example of Flow Alert through IFTTT Email:
image

.
.
### Example of Flow Alert MQTT Message format where "+" is the station index number:
opensprinkler/station/+/alert/flow{"flow_rate":5.00,"duration":60,"alert_setpoint":1.00}


booboodoguy (approved):
Very good changes.

booboodoguy:
I was hoping these changes would be included in your forked repository but unfortunately they are not. Could you please include them. Also since ifttt is no longer free could you finalize the mqtt additions.

bb09961:
The changes should be in my forked repository, make sure you choose the Flow_Alerts branch:
flow_alerts

booboodoguy:
I noticed in your current git hub Flow Alerts repository the following "comment code" "flow_gallons=0;"  is missing in:


`void turn_on_station(byte sid, ulong duration) {

	// RAH implementation of flow sensor
	flow_start=0;
	flow_gallons=0;
`


	`

bb09961:
I separated that into a separate pull request as I believe it is a bug that should be fixed even if the Flow Alerts Pull Request is not implemented.

Check Pull Request   #251  Added flow_gallons reset to Station Turn On

booboodoguy (approved):
I have been using this code for a few months now. But made changes to send email messages.

booboodoguy:
I have been using your implementation of OpenSprinkler on a Raspberry pi for a few months now. However, the key feature I need is for an excessive flow rate beyond the threshold set point  is to stop the station (zone) when this situation occurs. I have made changes to the code to send email messages for all the push notification messages instead of ifttt. I have been studying the code to determine how to add the "stop" upon flow rate error condition. Have you looked into this?

bb09961:
I'm glad its working for you, that's good to hear. 

So, instead of using IFTTT to send emails, you are sending them directly from a web client on the Raspberry pi?   That's better because IFTTT is now charging a subscription fee to use Webhooks to send emails for OpensSprinkler.

I would prefer that method as well, but I am using OpenSprinkler hardware and I don't know how to add an email server on the device.   I wish they would include that as a feature in the firmware.


Regarding the feature to stop a running station after a flow alert:    
Currently, the preexisting firmware logic that the flow alerts were built on calculates the flow rate after the station run is complete.    By using this logic, we are unable to stop the station because the alert is only triggered at the end of the run.   This was the least intrusive way to implement flow alerts in the existing firmware in hopes Ray would approve the pull request for everyone to use.


However, in the OpenSprinkler phone app, there is a feature where it shows the current flow rate in the bottom status bar.   This is a real-time rate that is calculated from the last 30ish seconds of water flow.  That value is not tied to a specific station, so it would be more complicated to access and constantly compare it against a set point for the running station without interrupting the main logic flow.     Also, you would need to think about spikes in flow rates due to initial air pockets and other fluctuations during the run.   I think other manufactures like Rachio had to implement some logic to prevent shutting down stations needlessly.

bb09961:
Also, can you post in your repository the changes to send emails directly from the Raspberry Pi?  I'd be interested in looking at them.

booboodoguy:
Included is my main.ccp and a shell script that actually sends the mail message. The shell script is activated by a cron schedule. This not the most elegant solution but it does work and never misses a message. Here are the changes I made. They are at the end of the Push Notification function just before the Logging functions. I made  a tempfs folder named ostemp to store the message files.

`		//char postBuffer[1500];
//		BufferFiller bf = ether_buffer;
//		bf.emit_p(PSTR("POST /trigger/sprinkler/with/key/$O HTTP/1.0\r\nAlarm Alert Message"
//						"Host: $S\r\n"
//						"Accept: */*\r\n"
//						"Content-Length: $D\r\n"
//						"Content-Type: Harvey/json\r\n\r\n$S"),
//						SOPT_IFTTT_KEY, DEFAULT_IFTTT_URL, strlen(postval), postval);

		//Create file name and location for  mail message
		strcat_P(postval, PSTR("\r\n"));	
		string s = to_string(type);
		string t = ( "sudo mv /ostemp/temporary /ostemp/alert" + s);
		const char* const rename = (t).c_str();					
		
		// Create file with the  postval buffer content for email message
		FILE *file;
		if (NULL==(file = fopen( "/ostemp/temporary" , "w"))) {return;}		
		fwrite(postval, 1, strlen(postval), file);
		fclose(file);		
		
		// Once the file is closed rename it to alert number			
		system(rename);					
`
[my_OS.zip](https://github.com/user-attachments/files/15945762/my_OS.zip)

booboodoguy:
Some thoughts on how to implement a stop station on excessive water flow.
1. In the push message "notify_station_on" function add the code to extract the station name flow rate number. Convert the number into gpm *4  as an integer, taking into account the adjustment value,  and  assign it to a variable "MaxRate"

2. In the void flow_poll function, right after the "flow_gallons==1" test, add code to capture the "flow_gallons" count during  4 minute periods. Test this value against the MaxRate and if excessive lock a "stop" state in a variable. This will repeat throughout the station run time.

3. In the "turn off a station" function, in the "// check if the current time is past the scheduled start time" code, add an additional test to see if the "stop"variable state is set to stop. Cause the station to turn off either upon time expiration or a "stop" state.

The 4 minutes could be less but the lower the number the more chance of a false alarm.
I know I'm missing several parameter conversions and some initialization but how does the logic sound to you?

booboodoguy:
One other thing to realize is that the running flow count in the app is not accurate as a result of several delays in the do_loop causing miss counts in the flow_poll routine when the app in connected. I have done extensive testing on this and found that the running flow count is not very useful when the app is running and the flow rate is around greater than 4hz. During my station run time the flow sensor frequency ranges from 7 to 15hz.

booboodoguy:
Are you going to merge your code with the new 2.2.1 release?
________________________________

bb09961:
Yes, once the new android app version is released, I'll dig into the firmware code and start testing the Flow Alerts.  I may need to make additional changes in order to incorporate the new Email Notifications. 

On a side note, I'm not sure why Issue #248 and corresponding pull request #251 wasn't incorporated into the new firmware.   
It looks like Ray fixed the flow rate bug you identified.   Let him know you also need pull request 251 for a reliable flow rate as well.

rayshobby:
Sorry we didn't have time to get this merged into firmware 2.2.1(0) but I will take a look at this for the next release. There are a lot of testing we had to do before releasing a firmware. We have to test it for AVR, ESP8266, Linux, Raspberry Pi, and in fact for Raspberry Pi itself we have to test it under all the cross combinations of hardware (Rpi 0, 4, 5) and Raspbian version (Bullseye, Bookworm) etc. So some PRs will take longer to review. Sorry about that, but it's on our list so it won't be missed.

bb09961:
I totally understand. Looks like you guys had lots of stuff to test and I'm really glad you were able to add in email notifications. That was a bigger priority for me as well.
 
I didn't see any comments acknowledging issue #248, so I thought the bug and PR was overlooked.  I'm glad to hear it's still on the list to be considered.

Thanks again.

booboodoguy:
Bb09961 I wonder if the new email implementation will be able to handle the three successive messages that occur when a station off happens in your flow alert fork. I had to queue the messages and email them out asynchronously.

 PUSH NOTIFICATION FUNCTIONS
case NOTIFY_STATION_OFF:
case NOTIFY_FLOW_ALERT:
case NOTIFY_FLOWSENSOR:

________________________________

bb09961:
Good point, I'll test that out.  I don't think I tested all 3 notifications enabled with IFTTT email either.

rayshobby:
It should be relatively easy to implement a simple email query to consolidate multiple messages into one email. Is there any reason to have IFTTT and email on both at the same time? The main purpose of email notification is to address the issue that IFTTT requires paid subscription now. If you already have IFTTT enabled, I don't see why you want to enable email as well separately.

bb09961:
No, I will not use IFTTT and the new Email feature at the same time.  I quit using IFTTT and started using MQTT when they started charging for webhooks.   But now that email notifications have been added to OpenSprinkler, I intend to switch over to that instead as it will simplify my setup.


My comment before was saying I haven't tested how each of the following alert methods will behave with multiple quick alerts triggered in succession:
IFTTT
MQTT 
New OpenSprinkler email feature

I'll test each method individually and report back what I observe.

bb09961:
I've merged my pull request with the lasted firmware.  The flow alerts can now be sent via Email, IFTTT or MQTT.

Also, I tested thee notifications types with all of three sending methods turned on at once and they were sent with no problems. 

All three of these were sent:
NOTIFY_STATION_OFF
NOTIFY_FLOW_ALERT
NOTIFY_FLOWSENSOR

using these three methods at the same time: 
Email
IFTTT
MQTT

booboodoguy:
Did you test using the email system on a raspberry pi?

bb09961:
No sorry. I only can test on the open sprikler 3.0 hardware.
-------- Original message --------From: booboodoguy ***@***.***> Date: 8/13/24  5:26 PM  (GMT-06:00) To: OpenSprinkler/OpenSprinkler-Firmware ***@***.***> Cc: bb09961 ***@***.***>, Author ***@***.***> Subject: Re: [OpenSprinkler/OpenSprinkler-Firmware] Flow Alerts (PR #252) 
Did you test using the email system on a raspberry pi?

booboodoguy:
What email service provider did you test the email performance on. I wonder if the email service will effect the probability of missing an email message.

bb09961:
I'm currently using gmail.   I tested it again today with the notifications enabled plus MQTT and I received all three email alerts within 15 seconds.

Wed, 14 Aug 2024 12:24:38
Wed, 14 Aug 2024 12:24:30
Wed, 14 Aug 2024 12:24:25

opensprinklershop:
Hi, I merged this pull request to the OpenSprinklerShop firmware and also added the "flow alerts" option to the notification options.
https://github.com/opensprinklershop/OpenSprinkler-Firmware
https://github.com/opensprinklershop/OpenSprinkler-App
Android app is also available here: www.opensprinklershop.de/firmware/app-release.apk

rayshobby:
Thanks for the contribution. This feature has been now integrated into firmware 2.2.1(1) code:
https://github.com/OpenSprinkler/OpenSprinkler-Firmware/commit/4ad81fd5c81bfebfdf9075e4fbdb400e7f28c368
A couple of minor issues:
1) the flow pulse rate should be calculated as the ((PULSE_RATE_1<<8)+PULSE_RATE_0) / 100.f. In your code I think you assumed PULSE_RATE_1 stores the integer part while _0 stores the fractional part:
https://github.com/bb09961/OpenSprinkler-Firmware/blob/Flow-Alerts/main.cpp#L1548
but this is not how these two bytes are calculated. I suppose this issue might not have come up because if the flow pulse rate is below 2.56 then byte _1 is always 0, so the two calculations are the same. But if the pulse rate is more than 2.56, then things will be different.

2) I discovered, painfully, that AVR's sscanf function doesn't support floating point variable, as a result, I had to switch sscanf to strtod. This is only relevant to OpenSprinkler 2.3 (which uses ATmega1284) and it doesn't affect OS 3.x (which uses ESP8266).

In any case, thanks again for the contribution!

bb09961:
Thanks for integrating this and for the detailed review.

You're right, I misunderstood the pulse rate variable and I didn't test it with a pulse rate factor larger than 2. Good catch.
And sorry about the sscanf frustrations. Looking at strtod now, it definitely seems like the better option.

Thanks again!

PR #254, opened 2023-10-03
Memory Leak during WiFi reconnect

AlexanderNikopol:
Every time the firmware reestablished the WiFi connection, HEAP memory decreased by more or less 5 kByte.
As a result the firmware stopped processing the station programs after the second or third reconnect. The firmware restarted automatically, when it was contacted via app or web GUI.

Problem was in function start_server_client. The call of this function adds the callbacks for the OpenThingsFramework. This function is called during each reconnect.
Because there is neither a function called, which removes the callbacks nor a check to prevent duplicate callback entries, the callbacks were added at the end of the callback list each time.

### Solution
Now an 'initialised' flag prevents the duplicate insertions.

### Hint
A better solution would be, to correct LinkedMapNode in the way, that it prevents the insertion of duplicate keys.
I did not do this, because I currently can not estimate the influences of such changes in this central object.

### DIFF
--- a/OpenSprinkler.h
+++ b/OpenSprinkler.h
@@ -66,6 +66,7 @@
 	#if defined(ESP8266)
 	extern ESP8266WebServer *update_server;
 	extern OTF::OpenThingsFramework *otf;
+	extern bool otf_callbacksInitialised;
 	extern ENC28J60lwIP eth;
 	#else
 	extern EthernetServer *m_server;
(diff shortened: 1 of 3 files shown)


opensprinklershop (approved):
Looks good.

repa6 (approved):

roelandkluit (approved):
Looks good

roelandkluit (approved):
Looks good

rayshobby:
Thanks for the fix. The changes are now in branch:
https://github.com/OpenSprinkler/OpenSprinkler-Firmware/commits/support_rpi_gpiod/
and will be merged when that branch is merged to master. You are right that ultimately the OTF library should be fixed to address this issue, and we are planning a few major changes to the OTF library soon.

Issue #264, opened 2024-04-10
Unusual DNS requests originating from opensprinkler


phoenix-9:
I was looking at tcpdump and noticed some unusual DNS queries originating from Opensprinkler:

11:16:51.637539 IP 192.168.17.19.61815 > 192.168.17.1.53: 1866+ A? M-DQ>^Cx^Ds^D^G^B3^E.dat. (34)
...
11:17:21.685907 IP 192.168.17.19.61815 > 192.168.17.1.53: 1867+ A? M-DQ>^Cx^Ds^D^G^BM-S^D.dat. (34)

These are invalid ("dat" is not a TLD), and probably indicate the presence of some other error in the code.

Firmware 2.2.0 (3)
Hardware 3.2 - AC

Let me know if you need more information.

arfrie22:
This should be fixed with #287, if this issue still persists after the next firmware update, 2.2.1. Please reopen the issue.

PR #287, opened 2024-07-24
Fix use-after-free when using custom ntp server

os04:
This also fixes issue #264 

Tested on my OpenSprinkler 3

**Issue:**

OpenSprinkler likes to periodically lookup gibberish DNS names when a custom NTP server has been configured. This lookup is often correlated with lookups of time.google.com and time.nist.gov, and the configured NTP server stops being queried when this starts to happen. Exact issue presentation can vary based on how some of the other settings are configured.

**Cause:**

It seems that configTime() requires its strings to be accessible after the function call is finished.

**Fix:**

Allocate a static buffer for the NTP server IP address string.

### DIFF
--- a/opensprinkler_server.cpp
+++ b/opensprinkler_server.cpp
@@ -2285,6 +2285,7 @@ void handle_web_request(char *p) {
 // othewise, using UDP is much faster for NTP sync
 ulong getNtpTime() {
 	static bool configured = false;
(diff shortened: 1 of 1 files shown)


arfrie22 (approved):

arfrie22:
Thank you for catching this.

PR #303, opened 2024-08-01
 I2C SSD1306 OLED Support and Button Support

arfrie22:
Add support for the OLED and buttons for the OSPI


arfrie22:
Merged into dev/221_1

PR #304, opened 2024-08-03
fix ospi startup script for otf

rayshobby:


rayshobby:
this didn't fix the root cause of the problem, so closing it

PR #323, opened 2024-12-14
chore(reorganize): reorganize project files

mellodev:
### Description
The goal of this PR is to clean up the project folder/file structure a bit to improve readability and maintainability. Moving the source files, header files and scripts out into dedicated folders seems like a good first step. This means lots of changes/refactors were made to `#include` statements, bash scripts and the github workflow yml. Lots of opportunity to accidentally break things here, so please review carefully!

### This PR:
- Moves all `.cpp`/`.c` files to `/src`
- Refactors `build.sh` g++ commands to pull in all files from `/src` rather than a hardcoded list
- Moves all `.h`/`.hpp` files to `/include`
- Refactors all `#include` statements to use new paths
- Moves all bash script files to `/scripts`
- Refactors various files which invoke scripts to use new path
- Updates SOURCES and HEADERS paths in `Makefile`

### Things that don't seem right
⚠️ `build.sh demo` does not  successfully compile on my end (but VSCode/Pio does). Could be environment specific, but please verify:
image

⚠️ `build.sh` g++ commands might need a little path tweaking for src and output folders
⚠️ `make all` also fails for me due to a WProgram.h missing include. Likely another environment specific issue?
⚠️ There were two changes made for `#include` statements for `WProgram.h` and `Arduino.h`. These can probably be reverted as they didn't resolve the `build.sh demo` failure I am having. VSCode/PlatformIO builds the project correctly either way.

image


rayshobby:

PR #325, opened 2024-12-22
Add a simple Docker Compose service definition.

developer79433:


developer79433:
This is failing CI build because it needs #326 .

PR #355, opened 2025-06-05
Fix real time flow to use a running average with dynamic timeout

### LINKED ISSUE #338: Feature request - no flow alert
I have been using the flow rate sensor code for a couple of years now and have experienced zero flow situations due to a defective rain sensor. Now that  the flow sensor code from bb09961 is in the main branch I request that a check be added to the flow sensor flow rate to, in addition to exceeding the user specified flow rate, a flow rate greater than zero ( or some small number) is checked.

arfrie22:
Fixes #338


rayshobby:
merged to dev/221_2

PR #324, opened 2024-12-22
Silence several compiler warnings

developer79433:

Issue #294, opened 2024-07-29
OpenSprinkler don't save anymore http hexadecimal characters in cmd, can't works anymore with Tasmota


Tikilou:
Hi! First of all, many thanks for all the work you've done on the opensprinkler.

With my previous opensprinkler setup under OsPI with electrical outlets connected under tasmota firmware, I was using http commands to turn them on and off. (To pump water from the well)
This release concerned firmware 2.2.0, running on Raspberry pi 4 and RaspiOS Bullseye.
Following an update to the OpenSprinkler firmware, I can no longer switch my connected sockets on and off.
The reason seems quite simple, visually at least I've noticed that when I indicate an http station type and register the start (cm?cmnd=Power%20On) or stop (cm?cmnd=Power%20Off of the socket) command, on last year's version 2.2.0, the command was registered as is.
Following the update, commands are now recorded as follows:

cm?cmnd=Power On
cm?cmnd=Power Off

As you can see, OpenSprinkler registers the command with a space without "%20", making it impossible to activate/stop the pump outlets.

To put this in context, I'm using OpenSprinkler (OsPI with Raspberry PI 4) with Home Assistant, Tasmota, Ecowitt, humidity and rain sensors, OpenWRT routers and Proxmox servers in a nursery I'm setting up.
I've written a big tutorial of my setup to popularize the use of OpenSprinkler and Open Source bricks in France for private and professional irrigation and water saving. :-)
https://pepinature.org/fr/gestion_de_l_eau/programmation_avancee_et_automatisation_de_l_irrigation

![Capture d’écran du 2024-07-29 15-04-37](https://github.com/user-attachments/assets/7252097b-54e0-4eea-a992-8ebcb08a9446)
![Capture d’écran du 2024-07-29 15-04-49](https://github.com/user-attachments/assets/04ba13a1-9b96-497a-8b9a-1568b89f79a8)

rayshobby:
The issue is that %20 is being URI decoded to a space character. I think you can solve this problem by URI encoding %, which is %25. So try:
cm?cmnd=Power%2520On
and see if it works. The only problem is that you have to do this encoding every time you save the HTTP zone settings, otherwise it gets translated to space again.

Tikilou:
[...]

PR #365, opened 2025-07-15
Analog rewrite

### LINKED ISSUE #397: [Feature] OTA update link on the web dashboard
### What feature do you want to see?

Link to the OTA update page can not be found anywhere on the web dashboard/UI. I know it's mapped on `/update` path according to the documents, but it will be great if this can be added somewhere on the web dashboard. Every time I need to check the documents on how to upgrade the firmware to remember how it was! If you add such a section to the settings, you can also add `check for new version` (along with a link to download) so that users don't have to navigate to an external website to download it.

### LINKED ISSUE #403: [Bug] GPIO not working on latest container image on OSPI (Trixie)
### Firmware Version

2.2.1-4

### Hardware Version

OSPI

### Bug Description

I upgraded my OSPI to Trixie, and upgrade the container image to the latest available tag (2.2.1-4), and re-imported the config backup created before upgrading.
This broke GPIO functionality, the GPIOs didn't respond at all no matter what I tried.

After this I installed the firmware (latest HEAD) and GPIOs are working again, with the same imported config.

### Steps to Reproduce

1. Install latest Raspberry Pi OS lite (based on Debian Trixie) on a OSPI.
2. Install Opensprinkler as a container following the README.Docker.md
3. Import config backup from previous, working container installation

### Errors

GPIOs don't work.

### LINKED ISSUE #415: fd leak in TinyWebsockets linuxTcpConnect → EMFILE hangs web UI & scheduler after network loss (OSPi)
## Summary

On the OSPi (Linux) build, every **failed** outbound WebSocket connection to the OTC cloud leaks one file descriptor. After a network outage the OTF WebSocket client keeps reconnecting (every 5 s), so the leaked descriptors accumulate until the process reaches `RLIMIT_NOFILE` (default 1024). From that point on, `accept()` on the local web-server socket fails with `EMFILE`, the web UI becomes unreachable, and the single-threaded main loop can no longer service requests — the controller looks "frozen" even though the process is still alive and was never rebooted. Scheduled programs are missed for the same reason.

## Environment

- Firmware `2.2.1(4)`, master, Raspberry Pi / OSPi (armhf, Linux)
- Submodules: `OpenThings-Framework-Firmware-Library` `4e93061`, `TinyWebsockets` `a939875`

## Root cause

`external/TinyWebsockets/tiny_websockets_lib/src/linux_tcp.cpp`, `linuxTcpConnect()`:

```cpp
int _socket = socket(AF_INET , SOCK_STREAM , 0);   // fd allocated
...
if ((he = gethostbyname(host.c_str())) == NULL) {
    return INVALID_SOCKET;            // leak #1: socket created, never closed
}
...
if (connect(_socket, ...) < 0) {
    return INVALID_SOCKET;            // leak #2: connect failed, never closed
}
return _socket;
```

Both error paths return `INVALID_SOCKET` after `socket()` succeeded, **without** `close(_socket)`. `LinuxTcpClient::connect()` then assigns the `-1` return value to `this->_socket`, so the real descriptor number is lost and can never be reclaimed by `close()`.

## Why it triggers in the field

- After a Wi-Fi drop the OTC cloud is unreachable. The OTF WebSocket client retries every `WEBSOCKET_RECONNECT_INTERVAL` (5 s, `OpenThingsFramework.cpp`).
- Each retry: `socket()` + `connect()` fails → one fd leaked.
- ~1000 leaks (~1.5 h) → the 1024 fd limit is reached → `accept()` → `EMFILE`.

## Evidence (from a controller caught in this state)

- Process alive, no reboot; **1001 socket fds** held, `FDSize 1024`.
- None of the 1001 socket inodes appear in `/proc/net/{tcp,tcp6,udp,...}` — i.e. fully-closed outbound attempts whose fd was leaked.
- `strace` of the hung main loop: a tight `poll(listen_fd)=POLLIN` → `accept() = -1 EMFILE` loop.

## Fix

Close the socket on both error paths before returning:

```diff
         if ( (he = gethostbyname( host.c_str() ) ) == NULL) {
           //"Failed to resolve hostname\n"
+          close(_socket); // avoid fd leak on DNS resolution failure
           return INVALID_SOCKET;
         }
@@
     if (connect(_socket, reinterpret_cast<sockaddr *>(&server), sizeof(server)) < 0) {
+      close(_socket); // avoid fd leak on failed connection
       return INVALID_SOCKET;
     }
```

After the fix, `strace` shows the descriptor being reused across attempts (`socket() = 28; connect() = -1 ETIMEDOUT; close(28) = 0`), and the process fd count stays flat under continuous connection failures.

## Note on the submodule

The affected file lives in the `gilmaimon/TinyWebsockets` submodule, which appears unmaintained (last commit 2020), so the change cannot be merged directly into this repo. Possible paths: carry a maintained fork of TinyWebsockets and bump the submodule pin, or apply the patch to the vendored copy. I'm happy to open a PR against whichever fork you'd prefer to track.

A separate, related weakness (not addressed here): the same `connect()` is blocking, so while the cloud is unreachable each reconnect stalls the single-threaded loop for the SYN timeout. A non-blocking connect with an explicit timeout would make reconnection fully transparent to the local web server.

### LINKED ISSUE #417: [Bug] Log filenames and end-timestamps default to 0/epoch due to format-string width mismatch in write_log() (RPi/Linux builds)
### Firmware Version

221

### Hardware Version

N/A

### Bug Description

On Raspberry Pi OS Lite (Trixie, 32-bit) running on a Pi Zero 2W, all station/program run logs were written to logs/0.txt regardless of actual date, instead of the expected logs/<day_number>.txt. Additionally, even after fixing the filename, the logged end-timestamp field for each run was 0, causing the web UI's table view to display run dates as Dec 31 1969 with start times calculated backward from midnight and end times shown as 00:00:00.
Root cause:
time_os_t (used for curr_time throughout main.cpp) resolves to time_t on non-Arduino builds (types.h). On this platform, time_t is 8 bytes (64-bit, signed) — confirmed via direct test — even though the OS itself is 32-bit (glibc made time_t 64-bit across the board for Y2038 mitigation). unsigned long on 32-bit ARM Linux is 4 bytes. Passing the 8-byte curr_time value to a %lu format specifier (expecting 4 bytes) causes snprintf to read only part of the value from varargs, producing 0 instead of the correct value.
Two call sites in write_log() (main.cpp) were affected:
1. Log filename generation:
csnprintf (tmp_buffer, TMP_BUFFER_SIZE, "%lu", curr_time / 86400);
2. End-timestamp field in the log line itself:
csnprintf(tmp_buffer + size, TMP_BUFFER_SIZE - size , "%lu", curr_time);
Verified via direct testing:
Added a debug print immediately before the filename snprintf call:
cfprintf(stderr, "DEBUG: curr_time=%ld write_log call\n", (long)curr_time);
Confirmed curr_time held a correct, sane Unix timestamp (e.g. 1782761040) at the point of the call — ruling out system time, NTP, or timezone misconfiguration. The bug is isolated to the format-string/type mismatch in the snprintf calls.
Fix:
Explicit cast to unsigned long at both call sites:
csnprintf (tmp_buffer, TMP_BUFFER_SIZE, "%lu", (unsigned long)(curr_time / 86400));
csnprintf(tmp_buffer + size, TMP_BUFFER_SIZE - size , "%lu", (unsigned long)curr_time);
After rebuilding with both changes, log files were correctly created with proper day-indexed names (e.g. 20633.txt), and the web UI table view correctly displayed actual run dates and times instead of Dec 31 1969.
Additional observation (low priority, not yet patched):
A related but distinct pattern exists elsewhere in opensprinkler_server.cpp, where os.now_tz() (return type time_os_t/time_t, 8 bytes) is assigned directly into variables explicitly declared as unsigned long (4 bytes), e.g.:
cunsigned long curr_time = os.now_tz();
(see lines ~771 and ~1813 in opensprinkler_server.cpp as of this writing). This causes a silent narrowing on assignment, but is currently harmless in practice because today's Unix timestamps (~1.78 billion) fit well within the 32-bit unsigned range (max ~4.29 billion). This would only become a real problem as timestamps approach the 32-bit unsigned overflow point (~2106), similar in spirit to the Y2038 problem that prompted time_t to become 64-bit in the first place. Flagging for awareness/future cleanup rather than as an urgent issue — no fix attempted or tested for this part.
Scope/platform note:
Both confirmed fixes are confined to the RPi/Linux (#else) branch of the relevant #if defined(ARDUINO) conditionals, so they do not affect Arduino-based builds, which use a separate time_os_t typedef (unsigned long) and a different file-handling path entirely. Not tested on 64-bit Raspberry Pi OS, but the same mismatch could plausibly occur there too if unsigned long remains narrower than time_t on that build.
Environment:

Hardware: Raspberry Pi Zero 2W
OS: Raspberry Pi OS Lite, Trixie, 32-bit
Kernel: 6.18.34+rpt-rpi-v7, armv7l
time_t size confirmed: 8 bytes, signed
Firmware version: 2.2.1 (build 221)
Build method: git clone --recurse-submodules + ./build.sh ospi

Note: Claude AI helped figure this all out

### Steps to Reproduce

On Opensprinkler Pi (with 32 bit lite OS on Pi Zero W2 - but likely on other PIs as well), run a station for a short while.  Try to look at the logs to see the event of that station run shows.  It does not.  Look at the log files.  The only one present is 0.txt rather than <number of days since the epoch>.txt.  Within that log file, you'll see that the last number in each row is 0, which is midnight on the morning of Jan 1, 1970 - the beginning of the epoch.

After fix, log files have he proper epoch number in front of .txt, and the entries in that log file have the proper non-zero number at the end of each line.

### Errors

No error messages.  Just lack of proper functionality as described above.

### LINKED ISSUE #418: [Feature] Sub zones
### What feature do you want to see?

I have a zone that currently goes to my garden, but there are various fixtures of different flows on it like drip lines, sprayers, etc. So I am going to add virtually controlled (via http) valves to further separate the garden zone (tree zone, drip zone, sprayer zone, etc). So with this I want to have each sub valve coupled with the master garden valve so if I turn on the tree zone, the garden zone will turn on first.

Need a delay offset value and configuration for if the master zone opens first or second, then if it closes first or closes second. You ideally want the master to close first to prevent hammering at the sub valve. Some delay would be good when opening, but I can't anticipate others use cases so building out flexibility would be easy.

arfrie22:
## Summary

This PR prepares firmware 2.2.1(5), a major update focused on support of the Sensor Expander and the Expanded Sensors API, while also improving scheduling, OSPi/Linux support, networking reliability, build tooling, and documentation.

  ## Expanded Sensors

  - Add a unified Expanded Sensor framework with stable UUID-based references.
  - Add ADS1115 analog sensors, including linear, piecewise-linear, and sensor-specific subtypes.
  - Add Aggregate, System Internal, and Onboard Digital sensor types.
  - Add sensor status and validity tracking, unit conversion, configurable logging, and program watering adjustments.
  - Add sensor metadata and API endpoints for configuration, readings, adjustment previews, and log access.
  - Add paginated sensor-log retrieval and deletion for responsive operation with large datasets and OTC connections.

  ## Controller and Scheduling

  - Add SN3 and SN4 support on compatible hardware.
  - Extend master support from two to four masters.
  - Allow effective runtimes after weather or sensor adjustment to exceed 16-bit and 18-hour limits, up to the seven-day runtime cap.
  - Correct repeated Run-Once program attribution in logs and last-run data.
  - Fix parallel-group scheduling bounds and station-index validation.

  ## OSPi/Linux and Networking

  - Replace platform-dependent time and duration types with fixed-width types, fixing logging on 32-bit Linux systems with 64-bit `time_t`.
  - Update and pin OTF 0.2.2 and the maintained TinyWebsockets fork.
  - Fix Linux WebSocket descriptor leaks and improve OTC/network recovery.
  - Restore OSPi GPIO initialization and correct ADS1115 I2C detection.
  - Build container images as OSPI rather than DEMO so GPIO support is present.
  - Improve firmware update scripts and deterministic submodule handling.

  ## Build, Security, and Maintenance

  - Fix memory/resource leaks, parsing bounds, timer rollover handling, and several API validation issues.
  - Refactor notifier, sensor, built-in sensor, and option definitions to reduce duplication and simplify future expansion.

  ## Documentation

  - Add the firmware 2.2.1(5) User Manual and API Reference.
  - Add dedicated Zone Expander and Sensor Expander manuals.
  - Add stable User Manual and API links.
  - Reorganize navigation and common firmware update/compilation documentation.
  - Document Expanded Sensors, four masters, SN3/SN4, longer effective runtimes, pagination, and the updated UI workflows.

  ## Issues Resolved

  Closes #417
  Closes #415
  Closes #403
  Closes #397
  Closes #418

  ## Superseded Pull Requests

  The following pull requests are superseded or incorporated by this PR and should be closed manually after merge:

  - #409, #406, and #383: Linux `time_t` and log-format fixes are covered by the fixed-width time-type changes.
  - #370: test-mode credential handling is implemented using an ignored local header and tracked example file.
  - #324: applicable compiler-warning fixes have been incorporated or made obsolete by the updated code.
  - #323: stale project-reorganization proposal; the exact directory layout proposed there was not adopted.


salbahra:
I found a duration-overflow issue while reviewing OpenSprinkler-App PR #294 against this firmware branch.

`RuntimeQueueStruct::dur` is `uint16_t` (`program.h:122`), but both adjusted-duration paths compute into `uint32_t` and assign without saturation:

- Scheduled programs: `main.cpp:824` assigns `water_time` after weather and sensor adjustment.
- Manual/program run path: `main.cpp:1649` assigns `dur` after the same adjustment.

For example, a valid 40,000-second duration at a 2.0 sensor factor becomes 80,000 and wraps to 14,464 when stored in the queue. Weather adjustment up to 250% can produce the same problem. App PR #294 now bounds App-originated Run-Once durations, but scheduled programs and other firmware/API paths remain affected.

Suggested fix: saturate the final adjusted duration to the queue maximum immediately before both assignments, while preserving the existing weather integer arithmetic and sensor-factor behavior. A regression should cover at least a result above 65,535 for both queue paths.

opensprinkler_server.cpp
```diff
@@ -1173,41 +1062,61 @@ void server_json_programs_main(OTF_PARAMS_DEF) {
...
+	begin_response(res);
+	print_header(OTF_PARAMS);
+	bfill.emit_p(PSTR("{\"jpa\":["));
+	ProgramStruct prog;
+	for(uint8_t pid=0; pid<pd.nprograms; pid++) {
+		pd.read(pid, &prog);
+		float wa = get_program_water_percent(prog) / 100.f;
+		float sa = get_program_sensor_adj(pid);
+		if(pid) bfill.emit_p(PSTR(","));
+		bfill.emit_p(PSTR("{\"wa\":$E,\"sa\":$E,\"ta\":$E}"), wa, sa, wa * sa);
```
salbahra, line 1123: **Expose the effective-runtime limit alongside `/jpa` factors**

`water_time_scale()` now caps each effective station runtime at `MAX_RUNTIME_DURATION` (604,800 seconds), but `/jpa` exposes only the uncapped `wa`, `sa`, and `ta` factors. [App PR #294 consumes these factors in its scheduled-program preview](https://github.com/OpenSprinkler/OpenSprinkler-App/pull/294#discussion_r3609177544) and currently computes `base x wa x sa` without the firmware cap, so it can display a longer run than the controller will execute. For example, 64,800 x 2.5 x 4 is 648,000 seconds in the preview versus 604,800 seconds in the queue. Since the limit is per station and cannot be represented by simply clamping `ta`, could `/jpa` expose the maximum effective runtime, such as a top-level `max_runtime`, or otherwise provide a stable API contract that the App can apply as `min(computed, max_runtime)`? This keeps preview and execution aligned without hard-coding a firmware-only constant in the App.

main.cpp
```diff
@@ -874,47 +782,39 @@ void do_loop()
 				bool will_delete = false;
 				unsigned char runcount = prog.check_match(curr_time, &will_delete);
 				if(runcount>0) {
+					const bool repeated_runonce =
+						strncmp_P(prog.name, PSTR(RUNONCE_REPEAT_PREFIX), sizeof(RUNONCE_REPEAT_PREFIX) - 1) == 0;
```
salbahra, line 787: **Do not classify run-once programs by a user-controlled name**

`/cp` accepts arbitrary program names, so an ordinary scheduled program whose name begins with `Run-Once with repeat` is classified here as run-once and receives `RUNONCE_PID`. `process_dynamic_events()` skips every queue PID at or above `MANUAL_PID`, which means that normal program can bypass controller-disabled, rain-delay, and per-sensor shutdown handling in addition to being misattributed in logs and notifications. Please identify generated repeated run-once programs through explicit internal metadata rather than a name prefix, and add a regression using an ordinary program with this prefix.

opensprinkler_server.cpp
```diff
@@ -1809,12 +1644,12 @@ void server_change_manual(OTF_PARAMS_DEF) {
...
 
-	uint16_t timer=0;
-	unsigned long curr_time = os.now_tz();
+	uint32_t timer=0;
+	uint32_t curr_time = os.now_tz();
 	if (en) { // if turning on a station, must provide timer
 		if (findKeyVal(FKV_SOURCE, tmp_buffer, TMP_BUFFER_SIZE, PSTR("t"), true)) {
-			timer=(uint16_t)atol(tmp_buffer);
-			if (timer==0 || timer>64800) {
+			timer=strtoul(tmp_buffer, NULL, 10);
```
salbahra: **Validate manual durations before narrowing to `uint32_t`**

On 64-bit Linux and OSPi, `strtoul("4294967297")` produces a 64-bit value that is narrowed to `1` in `timer` before the 18-hour bound is checked, so malformed oversized input is accepted. The same pattern exists in `mqtt.cpp` at the manual-run timer parse. Please parse into a wide type, validate the end pointer and range before casting, ideally through one shared helper, and regress both HTTP and MQTT inputs above `UINT32_MAX`.

opensprinkler_server.cpp
```diff
@@ -2077,78 +1880,865 @@ void server_pause_queue(OTF_PARAMS_DEF) {
...
+
+	// Iterate files from oldest to newest
+	uint16_t first_file  = hdr.wrapped ? (uint16_t)((hdr.cur_file + 1) % hdr.max_files) : 0;
+	uint16_t total_files = hdr.wrapped ? hdr.max_files : (uint16_t)(hdr.cur_file + 1);
+
+	SensorLogRecord rec;
+	uint32_t flat_idx = 0;
+	uint32_t count = 0;
+
+	for (uint16_t fi = 0; fi < total_files && count < max_count; fi++) {
```
salbahra: **Avoid scanning all retained history for an `after` query**

[App PR #294 now uses a one-day `after` cutoff for its initial chart load](https://github.com/OpenSprinkler/OpenSprinkler-App/pull/294#discussion_r3609177224), but this loop starts with the oldest file and only rejects pre-window records after reading each one. Linux and OSPi can retain 50 x 65,535 = 3,276,750 records, so even a one-day refresh may read nearly the entire log before returning the bounded response. Please use per-file timestamp bounds or newest-first traversal until the cutoff, while preserving response order, and add a behavioral check that a recent window does not read every retained file.

mqtt.cpp
```diff
@@ -206,7 +202,7 @@ void manualRun(char *message){
 				q->st = 0;
 				q->dur = timer;
 				q->sid = sid;
-				q->pid = 99;
+				q->pid = MANUAL_PID;
```
salbahra, line 204: **Reject Master 3 and Master 4 in MQTT manual runs**

HTTP `/cm` rejects all four configured master stations, but this MQTT manual-run path checks only `mas` and `mas2` before reaching this enqueue. A station configured as Master 3 or Master 4 can therefore be run independently through MQTT even though the equivalent HTTP request is rejected. Please use the same four-master predicate, preferably a shared helper, and add MQTT coverage for each configured master.

main.cpp
rayshobby: Programs created by users are not named like "Run-once with repeat". This is purely auto-created run-once with repeat programs.

opensprinkler_server.cpp
rayshobby: This has been addressed in the latest code.

mqtt.cpp
rayshobby: This has been addressed in the latest code.

opensprinkler_server.cpp
rayshobby: Addressed in 80aa7da. Queries using after now inspect each file’s newest record and skip the entire file when it predates the requested window, preserving chronological output without scanning all retained records. Also, Linux/OSPi retention was reduced to 50 × 16,384 records.

PR #370, opened 2025-08-08
Fix platformio to make it so testmode is ignored and generated at build time

arfrie22:


rayshobby:

PR #385, opened 2025-12-24
Update opensprinkler_server.cpp

HankBlardo:
Adds support to /jc endpoint to request specific parameters. In applications that cannot handle responses exceeding the MTU size, data not in the initial response packet are lost. By allowing selection of specific parameters, it ensures all data can be retrieved piecemeal.

ex. os-ip/jc?pw=a6d82bced638de3def1e9bbb4983225c&param=1&curr=1 will return the current value
// {"param":1,"curr":313}
ex. os-ip/jc?pw=a6d82bced638de3def1e9bbb4983225c&param=1&devt=1&ps=1 // {"param":1,"devt":1766315040,"ps":[[0,0,0,255],[0,0,0,255],[0,0,0,255],[0,0,0,255],[0,0,0,255],[0,0,0,255],[0,0,0,255],[0,0,0,255]]}

### DIFF
--- a/opensprinkler_server.cpp
+++ b/opensprinkler_server.cpp
@@ -1226,119 +1226,282 @@ function rst_wsp() {document.getElementById('wsp').value='$S';}</script>)"),
 void server_json_controller_main(OTF_PARAMS_DEF) {
 	unsigned char bid, sid;
 	time_os_t curr_time = os.now_tz();
-	bfill.emit_p(PSTR("\"devt\":$L,\"nbrd\":$D,\"en\":$D,\"sn1\":$D,\"sn2\":$D,\"rd\":$D,\"rdst\":$L,"
-										"\"sunrise\":$D,\"sunset\":$D,\"eip\":$L,\"lwc\":$L,\"lswc\":$L,"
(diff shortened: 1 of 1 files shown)


rayshobby:
Thanks for the PR. Could you clarify which specific client or application you are using that is experiencing this data loss?

Standard clients (Browsers, iOS/Android apps, Python scripts) handle TCP fragmentation automatically at the OS level, so MTU size should not cause data loss. We want to avoid adding complexity to the firmware API to solve issues that might be better fixed in the client's network handling.

HankBlardo:
It's a custom ARM-based datalogger. It's typically used to do industrial
MODBUS TCP calls, but can emit HTTP requests. Unfortunately it doesn't have
the sophistication to handle larger responses. I was able to satisfy my
needs by re-ordering the response to have the items of interest in the
first chunk and compiling my own firmware.  I thought it might be helpful
to generalize in case someone else has a similar constraint. I have no
problem with the pull request being denied/closed if adds unnecessary
complexity.

rayshobby:
OK I see. I suppose another approach is to create a separate endpoint that serves a 'compact' list of json variables - only those that you are interested in logging, and it's short enough to be still within the size limit of your datalogger. I am closing this PR for now because I feel the way it's currently implemented makes it cumbersome to modify the firmware in the future, and also because you already found a work-around to fit your need.

Issue #373, opened 2025-08-17
[Feature] Fertirrigation capabilities


LuisMMMTS:
### What feature do you want to see?

I want to be able to set a valve as a fertilizer valver and for each station in a program I would be able to set a time or percentage of time where the fertilizer valve is also open (it would ideally try to only open and close in the middle of the cycle so that there is a proper distribution of the fertilizer.

PR #386, opened 2025-12-24
Fertigation implementation

LuisMMMTS:
Relates to issue #373 
Define a station as fertilization valve
- for each station define either % or time of fertilization
- while master and station is running, fertilization valve is also open
- center fertilization in the middle of station run so that its active when pipes are already full and also stops before master valve so that water can push the remainds of fertilizer out of the pipes

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -64,6 +64,7 @@ unsigned char    OpenSprinkler::weather_update_flag;
 unsigned char OpenSprinkler::attrib_mas[MAX_NUM_BOARDS];
 unsigned char OpenSprinkler::attrib_igs[MAX_NUM_BOARDS];
 unsigned char OpenSprinkler::attrib_mas2[MAX_NUM_BOARDS];
+unsigned char OpenSprinkler::fert_station = 255;  // 255 = not configured
 unsigned char OpenSprinkler::attrib_igs2[MAX_NUM_BOARDS];
 unsigned char OpenSprinkler::attrib_igrd[MAX_NUM_BOARDS];
 unsigned char OpenSprinkler::attrib_dis[MAX_NUM_BOARDS];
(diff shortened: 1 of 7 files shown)


LuisMMMTS:
Is there anyone that can review this like @rayshobby or someone with privileges?

rayshobby:
Thanks for submitting this! This is a useful but specialized feature that addresses a specific use case.

Currently, our policy is to keep the core firmware focused on features we can fully support, understand, test, and document. Because this is a niche implementation, we are hesitant to merge it immediately as it would become a long-term maintenance liability for us.

For now, we're going to hold off on merging. If we see increased demand from other users for this specific functionality, we will revisit it. In the meantime, we’ll keep this PR open for those looking for this feature.

LuisMMMTS:
Thank you. Yes this probably is more for professional use. I have also made a pull request on the UI side, https://github.com/OpenSprinkler/OpenSprinkler-App/pull/284 please add the same comment there in case someone comes looking for this later on.

greenbaarron:
@LuisMMMTS , what are you using to inject the fertilizer? Are you doing anything w/ inground dripline?

joshlawless:
I would also like to see this implemented, rather than having to enable it through complicated program setup.  I am currently rebuilding my 12-zone residential landscaping and adding a fertigator (an EZ-FLO pressure-side tank with venturi inlet and outlet to be controlled by 24V solenoids) that injects fertilizer or insecticide.  I plan to control whether its open and injecting with a zone configured as a parallel zone, so I can set up programs to run each grass watering zone with the injector operating for a few minutes at a time (rather than during the irrigation cycle), and then closed for a few minutes to purge the lines.  Having UI support for this would be very helpful.

LuisMMMTS:
@greenbaarron I trigger a relay that drives a pump to inject fertilizer mixed with water into the system while the irrigation cycle is running. That's why I try to have the fertilization period evenly spaced between the beginning and the end of the cycle.

This ensures that the system is already full of water and properly pressurized, allowing the fertilizer to be distributed evenly. It also leaves some time at the end of the cycle to flush the system with clean water.

If I'm using 100% fertilization, then the pump runs from the start to the end of the cycle. currently its an above ground dripline that I use to water the field but I might upgrade it in the future to be inground

PR #406, opened 2026-05-11
Change format specifier for curr_time to unsigned long long

mkmer:
Since 221(4), logs have been placed in 0.txt file on my pi based controller. Correcting the type mismatch has restored the log file naming.


rayshobby:

PR #411, opened 2026-06-03
support d1 mini board

iofirag:
and show them as free pins in UI


rayshobby:
Thanks for the contribution, but these pin defines apply globally to all hardware revision 0 users —    
  disabling sensor pins 12 and 13 would break sensor functionality for everyone. This is a customization  
  that belongs in a personal fork rather than the main codebase.

iofirag:
Sure. It just for show what changes needed for build it to `d1 mini` board

You absolutly should close this PR :)

PR #420, opened 2026-08-08
Add pressure sensor

olesz:
Adding a possibility to select sensor type "pressure" as a binary output sensor


olesz:
Tested locally with the demo mode, can be selected, also status footer is properly populated when the sensor is activated.

Also see OpenSprinkler-App related pull request: https://github.com/OpenSprinkler/OpenSprinkler-App/pull/300

rayshobby:
Thanks for implementing and testing pressure-sensor support, including the matching App changes. Rain, soil, pressure, and other dry-contact sensors use the same binary-input behavior internally. Rather than add a separate firmware type for each application, we plan to introduce a generic binary sensor with an editable name. That will cover pressure switches and other use cases while keeping the existing Rain and Soil options for compatibility. I am closing this PR in favor of that more general approach; we appreciate the contribution.

PR #421, opened 2026-08-15
Implement check_network() watchdog for wired Ethernet (ESP8266)

cl445:
## Problem

On ESP8266 a wired controller has no recovery path at all. In WiFi mode the SDK reconnects on its own and `do_loop()` explicitly relies on that (`// WiFi disconnected, ESP8266 will handle re-connect`), but nothing re-establishes a wired connection: `start_ether()` runs exactly once, from `do_setup()`. That is why `docs/docs/troubleshooting.md` currently has to recommend setting up an auto-reboot program as the workaround for "connects initially but doesn't stay connected".

`check_network()` is the function meant to handle this. On ESP8266 it has been an empty stub since the 2.1.7 port (`7596d78`, 2017). The ping-and-reinitialize implementation it replaced was guarded by `#if defined(ARDUINO) && !defined(ESP8266)`, so the ESP8266 build never got one. Its doc comment still describes that older implementation.

The supporting scaffolding is all still in place and currently unused:

- `os.status.req_network` is raised every `CHECK_NETWORK_INTERVAL` in `do_loop()` and never read anywhere.
- `os.status.network_fails`, a 3-bit field sized for the old `1 << network_fails` backoff, is only ever written in `do_setup()`.
- `REBOOT_CAUSE_NETWORK_FAIL` was introduced in 2019 (`acbd2ca`) and used by the AVR variant of `check_network()`; it has had no user since AVR support was dropped in `39db235` (2025-11-17).
- `PING_TIMEOUT` in `main.cpp` is likewise unused.

## What this changes

`check_network()` is implemented for ESP8266, driven by the `req_network` flag that already exists.

It runs only when `useEth` is set, so WiFi behaviour is unchanged. That keeps the risk away from the vast majority of installations.

The check has two stages:

1. `eth.connected()` verifies link state and IP address. This catches an unplugged cable and an expired DHCP lease.
2. An ARP request to the gateway, collected two seconds later. This is what stage 1 cannot do: a stalled receive path leaves both link and address intact, so only an actual reply proves that frames still arrive. The gateway's cache entry is dropped via `etharp_cleanup_netif()` immediately before the request, because lwIP would otherwise keep a resolved entry for `ARP_MAXAGE` (300 s) and answer the lookup from cache, which would prove nothing.

After three consecutive failures (`NETWORK_FAILS_REBOOT`), which is 20 to 30 minutes at the existing 601 s interval depending on where the outage falls in the cycle, the controller reboots through the `safe_reboot` mechanism that is already there. It waits until no program is running and none is due within the next minute.

The watchdog keeps its own failure counter and only mirrors it into `os.status.network_fails` from the second consecutive failure onwards, so a single dropped ARP reply does not pause MQTT and the weather query (both bail out on that flag).

If some other reboot is already pending, the watchdog leaves it alone rather than taking over `safe_reboot`. Otherwise an unconditional `:>reboot_now` would be silently downgraded to an idle-only reboot.

On wired controllers `/db` gains three fields so the watchdog can be observed without waiting for a restart: `ethlink` (link and address per `eth.connected()`), `netfails` (consecutive failures) and `netreboot` (a reboot is armed). Without them the deferred-reboot behaviour cannot be told apart from the watchdog never having fired, which is what made the first hardware test unusable.

## Why reboot instead of re-initializing

`LwipIntfDev` in core 3.1.2 has no `end()`: `begin()` (`LwipIntfDev.h:179`) adds the netif via `netif_add()` (line 229) and registers a recurrent polling callback (line 276), and `config()` refuses to run once `_started` is set (line 149). A second `begin()` would add the same netif again and register a second polling callback. A chip-level reset below lwIP would be conceivable for the W5500 (`Wiznet5500::end()` exists) but not for the ENC28J60, which has none. A controlled reboot is the one recovery that works for both.

A controller that has already rebooted for a network failure will not reboot again until the connection has worked at least once, so an unplugged cable cannot put it into a reboot loop.

## Relation to #220

#220 addresses the same gap and I would rather not duplicate the intent behind it. This PR differs in scope deliberately: no new library dependency (#220 pulls in `bluemurder/esp8266-ping`), `DEBUG_` macros instead of `Serial.printf`, and nothing outside this one function.

## Testing

`pio run -e os3x_esp8266` builds clean. RAM 50.8 %, flash 58.5 %.

Verified on an OpenSprinkler 3.2 AC (hw_rev 2, ENC28J60) running over wired Ethernet, by pulling the cable and watching what the controller did:

| | |
|---|---|
| Cable pulled | 19:58:02 |
| Reboot | 20:21:36, i.e. 23 min later |
| Cable back in | 21:33:51 |
| `lrbtc` afterwards | 9 (`REBOOT_CAUSE_NETWORK_FAIL`) |
| `netfails` afterwards | 6 |
| `netreboot` afterwards | 0 |

23 minutes matches three failed checks at the existing 601 s interval, with the first one falling a few minutes after the cable came out. The reboot cause is one only this code sets.

The controller then ran for another 72 minutes with the cable still out, which is seven further check intervals. `netfails` shows the counter clamped at `NETWORK_FAILS_MAX` and `netreboot` shows no second reboot was armed, so the interlock against reboot loops does what it is meant to. Watering continued normally throughout the outage.

Before this, a first attempt at the same test could not be evaluated at all: a program was running for its whole duration, so `safe_reboot` correctly deferred the reboot, and from the outside that is indistinguishable from the watchdog never having fired. That is what the `/db` fields are for.

The `#else` branch for OSPi/Linux stays an empty function, so those targets are unaffected.

## Notes

`ARP_PROBE_WAIT` relies on `check_network()` being called once per second from `do_loop()`, which is also the cadence at which `req_network` is raised. Happy to switch this to a `millis()` comparison if you would rather not depend on that.

`etharp_cleanup_netif()` clears the whole ARP table of the wired interface, not just the gateway entry; lwIP exposes no narrower call. At one probe per 601 s the extra ARP traffic is negligible, but I am happy to revisit if you would rather not touch the table at all.

### DIFF
--- a/OpenSprinkler.h
+++ b/OpenSprinkler.h
@@ -108,6 +108,12 @@
 		inline wl_status_t status() {
 			return (isW5500)?w5500.status():enc28j60.status();
 		}
+		// the lwIP interface behind the active controller. LwipIntfDev hands it
+		// out as const, but etharp_request() takes a mutable netif, so the cast
+		// is kept here rather than at every call site.
+		inline netif* getNetIf() {
+			return (netif*)((isW5500)?w5500.getNetIf():enc28j60.getNetIf());
+		}
 	};
 	extern lwipEth eth;
 	extern bool useEth;
--- a/docs/docs/troubleshooting.md
+++ b/docs/docs/troubleshooting.md
@@ -239,7 +239,7 @@ If your controller can't connect or stay connected via wired Ethernet, try these
 2. **Disable PoE** (Power-over-Ethernet) on that port: OpenSprinkler’s wired module is **not PoE-compatible**. PoE-enabled port can cause it to malfunction.
 3. **DHCP Setting:** Leave the controller on DHCP, and set a **DHCP reservation** (IP-to-Mac) on your router to assign it a fixed IP. If you must use Static IP, verify **Gateway IP, DNS, Subnet Mask** are all correct and compatible with your router settings.
 4. **Use a dedicated router:** If your wired network is very busy, the most effective solution is to place OpenSprinkler on a dedicated router to isolate it from heavy traffic. 
-5. **Set an auto-reboot program:** If the controller connects initially but doesn't stay connected, a work-around is to set an auto-reboot program (daily or any regular interval of your choice). See [Program Name Annotation](manual.md#program-name-annotations) in the user manual.
+5. **Set an auto-reboot program:** If the controller connects initially but doesn't stay connected, a work-around is to set an auto-reboot program (daily or any regular interval of your choice). See [Program Name Annotation](manual.md#program-name-annotations) in the user manual. On firmware versions that include the wired connection watchdog this is usually unnecessary: the controller checks the connection about every 10 minutes and reboots itself if it stays unreachable, waiting until no program is running.
 
 **<u>For OSPi</u>:** Networking is managed by RPi. Ensure RPi has adequate power (OSPi supplies up to 500 mA. RPi 3/4/5 may need extra power via USB, otherwise network may be unstable).
 
--- a/main.cpp
+++ b/main.cpp
@@ -1890,32 +1890,49 @@ void delete_log(char *name) {
 /** Record the outcome of a network check, and reboot if the connection stays down */
 static void network_check_result(bool ok) {
 	/* Whether an outage may still trigger a reboot. Initialized on first use,
-	 * which is safely after os.begin() has restored last_reboot_cause: if the
-	 * previous boot already ended in a network reboot and the connection has not
-	 * worked since, rebooting again will not fix it either - the cable is simply
-	 * unplugged - so the controller stops trying. */
+	 * which is safely after os.options_setup() has restored last_reboot_cause:
+	 * if the previous boot already ended in a network reboot and the connection
+	 * has not worked since, rebooting again will not fix it either - the cable is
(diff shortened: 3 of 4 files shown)


cl445:
The red X on this PR and on #422 is not a build failure. The runs from 15 August needed maintainer approval, never got it, and GitHub expired them on 14 September with zero jobs. Could you approve a fresh run?

I also saw dev/221_6-architecture. check_network() is still a stub there, so this change still applies. It merges onto that branch with one conflict in main.cpp. Say the word and I'll rebase.

rayshobby:
Thanks for the careful investigation and for testing the ENC28J60 path on hardware. Also noted that the red CI checks expired without maintainer approval; they are not evidence of a build failure.

For 2.2.1(6), we have taken a different approach for ESP8266/W5500: the driver now bounds SPI and command waits, validates MACRAW frames before consuming them, detects stuck receive state, and first attempts a chip-level reinitialization. A controlled reboot is a last resort after repeated recovery failures. This targets controller faults directly and usually avoids rebooting because a gateway ARP probe failed. It is not identical coverage to your proposal, and it does not cover ENC28J60.

ENC28J60 is discontinued hardware for us, so we are prioritizing W5500 reliability for this release and are not rebasing this PR into 2.2.1(6). We appreciate the ENC28J60 testing and will leave this open for now while the 2.2.1(6) work is merged.

PR #422, opened 2026-08-15
Check the wired link before handing the interface to lwIP

cl445:
Three places where the wired connection is not handled the way the WiFi one is.

## `save_wifi_ip()` never stored a wired address

The function carried a `// todo: handle wired ethernet` and only ever ran for WiFi, so on a wired controller the stored address options stayed at whatever `start_ether()` wrote once at boot. If DHCP hands out a different address later, the values in Edit Options drift away from reality. It now takes the values from whichever interface is actually in use.

## `/jc` reported an RSSI on wired controllers

`useEth` implies `WiFi.mode(WIFI_OFF)`, so the reported value is whatever the radio last left behind. The field is now omitted on a wired connection rather than filled with a meaningless number. Clients that read it should already tolerate its absence, since it does not exist on OSPi builds either, but this is an API change worth pointing out.

## The wired interface could stay up after a fallback to WiFi

`start_ether()` called `eth.begin()` before it knew whether a cable was attached. That call registers the netif with lwIP, starts its DHCP client and installs the packet polling callback, and none of it can be taken back: `LwipIntfDev` in core 3.1.2 has no `end()`, and `config()` refuses to run once `_started` is set.

So when the link then failed and `fwire` was 0, the controller moved on to WiFi while the wired interface kept running. If a lease arrived after the timeout it answered on that address too, with `useEth` saying it was a WiFi device. I hit this on an OS 3.2 with an ENC28J60: reachable on two addresses at once, one per interface, with only the WiFi one reported in `/db` and on the LCD.

The fix is to ask the chip before involving lwIP. `LwipIntfDev` inherits publicly from the driver, so the raw driver is reachable:

```c
if(!eth.rawBegin(mac)) return 0;                    // chip only, lwIP untouched
while(!eth.rawLinked() && !timeout) delay(250);     // link straight from the PHY
if(!eth.rawLinked() && !fwire) return 0;            // clean fallback

if(!eth.begin(mac)) return 0;                       // only now: netif + DHCP
```

`rawBegin()` runs `ENC28J60::begin()` or `Wiznet5500::begin()`, `rawLinked()` reads `isLinked()`. Both are public on the driver classes, so this needs no change to the core.

Both settings keep an honest meaning. `fwire=1` waits for the link and takes the interface either way. `fwire=0` falls back to WiFi without ever creating a second interface.

One case remains: link up, but DHCP failing, with `fwire=0`. The netif is registered by then and cannot be removed, so `eth_started` still records it and `/db` reports the wired address under `ethip`. That is now a rare corner rather than the normal path.

Startup also gets shorter. The old loop waited up to 60 s for `eth.connected()`, which is link and address together. The link alone now gets `ETHER_LINK_TIMEOUT` (8 s) and DHCP keeps the remainder.

## Testing

`pio run -e os3x_esp8266` builds clean, RAM 50.8 %, flash 58.5 %.

Both paths were checked on an OS 3.2 AC with an ENC28J60.

**Cable attached, `fwire=1`.** The controller comes up on the wired interface as before, so the added link check does not get in the way of the normal path. `/jc` no longer carries an RSSI field, and the stored address options follow the wired DHCP lease.

**No cable, `fwire=0`.** The controller falls back to WiFi as intended. The point of the change is what happens next: after plugging the cable back in, the wired address stays silent.

```
$ ping 192.168.1.90
3 packets transmitted, 0 packets received, 100.0% packet loss

$ arp -n 192.168.1.90
? (192.168.1.90) at (incomplete) on en0

$ # full ARP sweep of the subnet: the wired MAC does not appear at all
```

`incomplete` means nobody answered the ARP request, and the interface's MAC is absent from the subnet, so the interface is not merely without an address, it is inactive. Before this change the same sequence left that address answering on its own DHCP lease while `useEth` reported a WiFi device, which is how the problem was found.

Related to #421, which adds the watchdog that recovers a wired connection once it is lost. The two do not overlap in code.

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -461,6 +461,22 @@ unsigned char OpenSprinkler::start_ether() {
 	}
 
 	load_hardware_mac((uint8_t*)tmp_buffer, true);
+
+	/* Bring up the chip on its own first and wait for the link there. eth.begin()
+	 * registers the netif with lwIP, starts its DHCP client and installs the
+	 * packet polling callback, and none of that can be taken back: LwipIntfDev
+	 * has no end(). Committing to it before we know a cable is attached is what
+	 * used to leave a second, invisible interface running after a fallback. */
+	if(!eth.rawBegin((uint8_t*)tmp_buffer)) return 0;
+	lcd_print_line_clear_pgm(PSTR("Wait for link"), 1);
+	uint32_t linkout = millis()+ETHER_LINK_TIMEOUT;
+	while(!eth.rawLinked() && (int32_t)((uint32_t)millis()-linkout)<0) { delay(250); }
+	if(!eth.rawLinked() && !iopts[IOPT_FORCE_WIRED]) {
+		// no cable and the user has not asked us to insist: leave lwIP untouched
+		DEBUG_PRINTLN(F("no wired link, falling back to WiFi"));
+		return 0;
+	}
+
 	if (iopts[IOPT_USE_DHCP]==0) { // config static IP before calling eth.begin
 		IPAddress staticip(iopts+IOPT_STATIC_IP1);
 		IPAddress gateway(iopts+IOPT_GATEWAY_IP1);
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -470,6 +470,7 @@ unsigned char OpenSprinkler::start_ether() {
 	}
 	eth.setDefault();
 	if(!eth.begin((uint8_t*)tmp_buffer))	return 0;
+	eth_started = true;
 	lcd_print_line_clear_pgm(PSTR("Start wired link"), 1);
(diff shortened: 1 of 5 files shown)


cl445:
Same as #421: the CI run from 15 August expired unapproved on 14 September, which shows as a failure with zero jobs.

Against dev/221_6-architecture this has three small conflicts, and the link check is still missing there. Happy to rebase if you prefer that target.

rayshobby:
Thanks for separating these three wired-network issues and for testing the fallback case on ENC28J60 hardware. I also understand the red CI checks expired before maintainer approval rather than failing a build.

We agree the live wired interface after fallback to WiFi is a real defect. The saved DHCP address being stale in Edit Options and the meaningless wired RSSI are worth cleaning up, but neither is a 2.2.1(6) release blocker; omitting RSSI also changes the /jc response shape. Since this release is close to completion, we are deferring this PR rather than introducing another network-startup change now. We plan to address the wired-interface fallback and revisit the other two items for 2.2.1(7), with a rebase and regression tests on the current architecture branch. Please do not spend time rebasing it for 2.2.1(6). We will leave the PR open for now.

Issue #423, opened 2026-08-18
[Bug] /co accepts brace-wrapped values for nested options (wto/mqtt/otc/email), making /ja emit invalid JSON


jheddings:
### Firmware Version

2.2.1(4)

### Hardware Version

v3.4 AC

### Bug Description

The `/co` endpoint stores the value of the nested-object options (`wto`, `mqtt`, `otc`, `email`) exactly as supplied, while the JSON output adds the enclosing braces itself. If a client sends a complete JSON object — braces included — those braces are stored as part of the value, and the output then carries two sets. The result is that `/ja` emits invalid JSON and can no longer be parsed.

The expected convention looks like "send the object body without the outer braces", and the firmware is entirely self-consistent about it. `/ja` emits `"wto":{$O}` at [L1201](https://github.com/OpenSprinkler/OpenSprinkler-Firmware/blob/6ac3b11066a945bb1b76aafe898da2e71cb66e41/opensprinkler_server.cpp#L1201), `"otc":{$O}` at [L1190](https://github.com/OpenSprinkler/OpenSprinkler-Firmware/blob/6ac3b11066a945bb1b76aafe898da2e71cb66e41/opensprinkler_server.cpp#L1190) and `"email":{$O}` at [L1213](https://github.com/OpenSprinkler/OpenSprinkler-Firmware/blob/6ac3b11066a945bb1b76aafe898da2e71cb66e41/opensprinkler_server.cpp#L1213), and on the write side the `/co` handler loads the stored value into `tmp_buffer+1` specifically to [`// make room for the leading '{'`](https://github.com/OpenSprinkler/OpenSprinkler-Firmware/blob/6ac3b11066a945bb1b76aafe898da2e71cb66e41/opensprinkler_server.cpp#L1491) before parsing it. The official UI follows this convention, so the behaviour only surfaces for third-party API clients.

The reason I'm raising it is that the failure is silent and a little hard to trace. Sending well-formed JSON is a natural thing for an API client to attempt; `/co` accepts it, returns `{"result":1}`, and — worth noting — the option values themselves are applied correctly. Only the serialisation of later `/ja` responses is affected. In my case a nightly config backup polls `/ja`, so the first sign of trouble was a stored snapshot that would not parse, well after the change that caused it.

[...]

PR #427, opened 2026-08-23
Accept brace-wrapped values for nested string options

jheddings:
Addresses #423.

## What this changes

The nested-object string options (`wto`, `mqtt`, `otc`, `email`) are stored without their enclosing braces — the emitters print them as `"key":{$O}`, `parse_wto()` writes the braces back before deserializing, and `server_change_options()` loads into `tmp_buffer+1` specifically to reserve room for the leading brace. A client that sends a complete JSON object therefore has its braces stored as part of the value, and `/ja` then emits `"wto":{{...}}`, which no longer parses.

This adds `strStripOuterBraces()` alongside the other string helpers in `utils.cpp` and calls it at the four `/co` save sites, so either encoding is accepted.

## Why it should be safe

Values written in the documented form are untouched: an object body begins with a quote, never a brace, so the strip is a no-op for existing callers. Only one matching outer pair is removed, so a nested object inside the value survives. The operation only ever shortens the string, so there is no buffer risk — and it frees the two spare bytes `parse_wto()` needs when it appends its closing brace.

## Testing

Built the OSPI target in Docker and exercised `/co` against a running instance, comparing an unpatched build with a patched one on the same inputs:

| Input | Before | After |
| --- | --- | --- |
| `wto={"h":20,"t":50}` | `"wto":{{"h":20,"t":50}}`, `/ja` fails to parse | parses, `{"h":20,"t":50}` |
| `wto="h":30,"t":60` | parses | parses, unchanged |
| `wto={}` | `"wto":{{}}`, fails to parse | parses, `{}` |
| `mqtt={"en":1,"host":"a"}` | fails to parse | parses, `{"en":1,"host":"a"}` |
| `mqtt="en":0,"host":"b"` | parses | parses, unchanged |
| `mqtt={"en":1,"x":{"y":2}}` | — | parses, inner object intact |

I have only exercised the OSPI/Linux build; I have not flashed this to an AVR or ESP target.

## One design question

I took the lenient route — accept either form — on the assumption that silently repairing existing broken clients is preferable. Rejecting brace-wrapped input with an error would surface the mistake more loudly and is a small change to the same helper. Happy to switch if you would rather it be strict, or to close this in favour of a documentation note if that is the direction you prefer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/opensprinkler_server.cpp
+++ b/opensprinkler_server.cpp
@@ -1487,6 +1487,7 @@ void server_change_options(OTF_PARAMS_DEF)
 	}
 	uint8_t keyfound = 0;
 	if(findKeyVal(FKV_SOURCE, tmp_buffer, TMP_BUFFER_SIZE, PSTR("wto"), true)) {
+		strStripOuterBraces(tmp_buffer);
 		if (os.sopt_save(SOPT_WEATHER_OPTS, tmp_buffer)) {
 			os.sopt_load(SOPT_WEATHER_OPTS, tmp_buffer+1); // make room for the leading '{'
 			parse_wto(tmp_buffer); // parse wto
@@ -1505,6 +1506,7 @@ void server_change_options(OTF_PARAMS_DEF)
 
 	keyfound = 0;
 	if (findKeyVal(FKV_SOURCE, tmp_buffer, TMP_BUFFER_SIZE, PSTR("otc"), true, &keyfound)) {
+		strStripOuterBraces(tmp_buffer);
 		os.sopt_save(SOPT_OTC_OPTS, tmp_buffer);
 	} else if (keyfound) {
 		tmp_buffer[0]=0;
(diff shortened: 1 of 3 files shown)


rayshobby:
Thanks for identifying the nested-object serialization bug and for the compatibility-minded fix. The upcoming 2.2.1(6) branch now accepts both the legacy object-body form and a complete brace-wrapped JSON object for wto, mqtt, otc, and email. This was implemented in commit 074bd57 (accept complete JSON objects in option updates), with native, Demo API, and ESP8266 hardware coverage.

We are not merging this older PR directly, but its reported behavior is addressed in the release branch. I will leave it open until the 2.2.1(6) PR is merged, then close it with credit to your contribution.

rayshobby:
Thanks, @jheddings, for reporting #423 and proposing the compatibility-minded fix. Merged PR #435 addresses it in commit 074bd57: /co accepts both legacy object-body values and complete brace-wrapped JSON objects for wto, mqtt, otc, and email. We implemented and tested the fix on the release branch, so this alternative PR is now closed with credit for your contribution.

Issue #432, opened 2026-09-06
[Bug] Watering level does not fall back to 100% when the weather service is unreachable


BrentIO:
### Firmware Version

2.2.1(5) (`OS_FW_VERSION 221`, `OS_FW_MINOR 5`)

### Hardware Version

OpenSprinkler 3.x (ESP8266). The logic is in shared code, so OSPi is affected too.

### Bug Description

When an automatic weather method (Zimmerman / ETo) cannot reach the weather
service, the firmware keeps applying the last `scale` it received - with no time
limit. If the last value was a `0%` rain adjustment, every weather-enabled
program run is silently skipped. If it was an above-100% heat adjustment, every
run over-waters. Either way the adjustment outlives the conditions that produced
it.

`check_weather()` has a watchdog (`main.cpp:1119`) meant to reset the watering
level to 100% after `CHECK_WEATHER_SUCCESS_TIMEOUT` (24 h) with no successful
weather call. It does not work in the cases that matter:

1. It is gated on `os.checkwt_success_lasttime`, which is **RAM-only** - `0`
   after every reboot, and only ever set on an `errCode==0` response
   (`weather.cpp:68`). So after a reboot during an outage - or on a controller
   hitting a persistent `errCode`, or one that has never reached the service -
   the guard is never true. `IOPT_WATER_PERCENTAGE` **is** persisted to flash
   (`weather.cpp:77`), so the stale value survives the reboot while the
   mechanism meant to clear it does not. Watering stays at the stale value
   indefinitely.
2. It sits **behind** the `network_fails` / `network_connected()` early-returns
   (`main.cpp:1113`-`1116`), so during the network outage that usually *causes*
   the problem, the watchdog never runs at all.

**Expected:** when the firmware can no longer confirm that the weather
adjustment is current, it should return to 100%, and that recovery should
survive reboots. Manual / AutoRainDelay / Monthly methods and non-weather
programs should be unaffected.

### Steps to Reproduce

1. Firmware 2.2.1(5), weather method = Zimmerman or ETo.
2. Have the weather service return `errCode=0, scale=0` once, so `wl` becomes 0,
   is saved to flash, and `checkwt_success_lasttime` is set.
3. Make the weather host unreachable.
4. Reboot the controller (`rbt` or power-cycle).
5. `jc` shows `lupt` (`checkwt_success_lasttime`) = `0`, `wl` = `0`, and it
[...]

PR #433, opened 2026-09-06
Reset watering level to 100% when weather data is stale

BrentIO:
check_weather() already resets the watering level to 100% when there has been no successful weather call for CHECK_WEATHER_SUCCESS_TIMEOUT (24h), so a stale weather adjustment - a 0% rain adjustment, or an above-100% heat adjustment - is not carried forward forever when the weather service is unreachable. That reset does not work in the cases that matter:

- It is gated on os.checkwt_success_lasttime, which is RAM-only and 0 after every reboot, and is only set on an errCode==0 response. So after a reboot during an outage (or a persistent errCode, or a controller that has never reached the service) the guard is never true. IOPT_WATER_PERCENTAGE *is* persisted, so the stale value survives while the mechanism to clear it does not.
- It sits behind the network_fails / network_connected() early-returns, so it never runs during the network outage that usually causes the problem.

Changes:

- NVConData gains weather_success_lasttime; getweather_callback() writes it (rate-limited to hourly, piggybacking the existing nvdata save) on every errCode==0 response, and nvdata_load() seeds checkwt_success_lasttime from it on boot. The 24h clock is now measured from the last actual contact, so a reboot - or a reboot loop - does not restart it. An old, shorter nvdata file reads the new field back as 0, which falls through to the powerup_lasttime fallback (same as before).

- check_weather(): the reset uses powerup_lasttime as the reference when there has been no successful call since boot, runs before the network / remote-extension early-returns so it works while offline, persists the reset via iopts_save() (guarded on an actual change), and sets WEATHER_UPDATE_WL so the user is notified. The program_busy guard stays first, so no flash write happens mid-program.

Manual / AutoRainDelay / Monthly methods and non-weather programs are unaffected. CHECK_WEATHER_SUCCESS_TIMEOUT (24h) is unchanged.

### DIFF
--- a/OpenSprinkler.cpp
+++ b/OpenSprinkler.cpp
@@ -2147,6 +2147,10 @@ void OpenSprinkler::options_setup() {
 /** Load non-volatile controller status data from file */
 void OpenSprinkler::nvdata_load() {
 	file_read_block(NVCON_FILENAME, &nvdata, 0, sizeof(NVConData));
+	// seed the in-RAM last-success time from nvdata so the stale-weather watchdog
+	// in check_weather() is not reset by a reboot (an old, shorter nvdata file
+	// reads this field back as 0, which is handled)
+	checkwt_success_lasttime = nvdata.weather_success_lasttime;
 	old_status = status;
 }
 
--- a/OpenSprinkler.h
+++ b/OpenSprinkler.h
@@ -125,6 +125,7 @@ struct NVConData {
 	uint32_t external_ip;        // external ip
 	uint8_t  reboot_cause;       // reboot cause
 	uint16_t last_sensor_uuid;   // counter for sensor UUID generation; next sensor gets ++this
+	uint32_t weather_success_lasttime; // last time the weather service returned errCode==0 (0 = never)
 };
 
 struct StationAttrib {  // station attributes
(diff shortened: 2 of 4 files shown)


rayshobby:
Thanks for reporting the stale weather-adjustment failure and for proposing a fix. The problem is addressed on the upcoming 2.2.1(6) branch in commit 4a7eba5 (fail safe when weather data becomes stale), though with a different implementation from this PR.

Rather than persist a last-success timestamp, the firmware now checks response freshness where weather-derived values are consumed. After a reboot or once a successful response is more than 24 hours old, Zimmerman/ETo runs use a neutral 100% watering level and stale weather restrictions no longer block runs. A later maintenance pass also reconciles the stored/displayed state, even when the network is unavailable. Manual watering levels are preserved. This avoids making the fail-safe depend on a new persisted timestamp and covers the reboot/outage cases you identified.

Native and Demo API tests cover the policy and integration paths. We will leave this PR open until the 2.2.1(6) PR is merged, then close it with credit for the issue and proposed fix.

rayshobby:
Thanks, @BrentIO, for identifying the stale weather-adjustment failure and proposing a fix. Merged PR #435 addresses #432 in commit 4a7eba5 using point-of-consumption freshness checks rather than a persisted timestamp. After 24 hours without a successful response, weather-derived scaling falls back to 100% and stale restrictions expire, including across reboot or network loss. Closing this alternative PR with credit for the report and proposed fix.

